PR #1, opened 2015-09-14
Promnet

qed-:
This change allows specifying an explicit hostname to MetricServer. The use case is to pass 'localhost' to allow the embedded HttpListener to run without administrator privileges.

Also changed the collector registry to use ConcurrentDictionary rather than explicit locks.


qed-:
Any chance this could be pushed up to Nuget?

andrasm:
sure, v0.0.5 is now available on nuget.

PR #6, opened 2015-11-26
Default collector changes

qed-:
Allow specifying alternative default collectors (or none).
Implement some process/.net metrics that do not rely on performance counters.

My main use case is to avoid relying on performance counters for metrics, which has proven problematic in practice on some of our servers. It seems this is a common problem: http://ayende.com/blog/165217/performance-counters-sucks

I have written an alternative default collector and made it possible to use this instead. By default the existing perf counter based metrics will continue to be used.


andrasm:
thanks a lot, will be available on nuget soon (v0.0.7)

PR #7, opened 2015-12-01
Add interfaces

qed-:
Adds ICounter, IGauge, IHistogram, ISummary implemented by both parent and child objects.
Also IMetricServer for completeness.

### DIFF
--- a/prometheus-net/Counter.cs
+++ b/prometheus-net/Counter.cs
@@ -4,7 +4,13 @@ using Prometheus.Advanced.DataContracts;
 
 namespace Prometheus
 {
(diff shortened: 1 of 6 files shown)


andrasm:
thanks - will be available on nuget soon.

qed-:
I think something's up with Nuget, the DLL is stamped 0.0.5.0 and doesnt have the DotNetStatsCollector class

andrasm:
?

just tried it on a fresh project - it shows v0.0.8 with DotNetStatsCollector + the new interfaces available

qed-:
Must have been a problem with my project, working now.

PR #8, opened 2015-12-04
Generalise scraping so it can be called externally without using the embedded http handler

qed-:
The use case here is to allow the client to be embedded in web apps that have their own http handlers, without having to open up a separate port.

To do this the MetricServer.Start method would not be called, instead ProcessScrapeRequest would be called by the hosting framework when the /metrics path is requested.


andrasm:
thanks - on nuget as v0.0.9

PR #11, opened 2015-12-30
Public ScrapeHandler

qed-:


qed-:
Hi,

Thanks for giving me pull access on the repo. I'll keep doing things via
pull requests if I make any significant changes so you can review it.

How does the nuget part work? Would you mind uploading my summary changes
there?

Thanks,
David

andrasm:
Hi David - it should be on nuget now (v0.0.11). I think we can call it 1.0
soon - anything you wanted to add before that?

thanks,
andras

qed-:
Hi Andras, my only suggestion before v1 would be to make the non-perf
counter metrics the default.

Cheers,
David

PR #14, opened 2016-02-22
Change some metric names

qed-:
Following discussion of f41d9d9


andrasm:
Sounds like this qualifies as a breaking change, though it would be a shame bumping the major version just for this. What about inactivating the current 1.0.0 package from nuget and release these changes as 1.1.0?

qed-:
Fine by me.

qed-:
Are you happy to integrate this @andrasm?

andrasm:
right - merged + pushed to nuget as v1.1

PR #17, opened 2016-05-19
Use UTF-8 encoding for Text output

juergenhoetzel:
I had issues using non-ascii chars in help text and label values. This is fixed by setting the output encoding to _UTF-8_ as  described here: https://prometheus.io/docs/instrumenting/exposition_formats/


andrasm:
Thanks for the pull request. Pushed the fix to nuget as version 1.1.2

PR #18, opened 2016-05-26
Add option for setting up endpoint as https.

mattias01:


andrasm:
thanks, pushed to nuget as v1.1.3

mattias01:
Thanks for prometheus-net.

PR #21, opened 2016-07-02
replaced lock with Interlocked operation where aplicable (Counter, Ga…

MihaMarkic:
Replaced lock statement in Counter and Gauge. Per. improvements on my computer are like this:

              Method |     Median |    StdDev |
-------------------- |----------- |---------- |
            LockTest | 19.5176 ns | 1.6874 ns |
        LockReadTest | 19.1564 ns | 0.0337 ns |
     InterlockedTest | 13.7282 ns | 0.0290 ns |
 InterlockedReadTest |  9.3791 ns | 0.1556 ns |

Reading value is almost twice faster and writing is considerably faster as well. Couldn't test on .netcore/linux since benchmark library isn't ported there yet.
And fixed a bug in Gauge where Interlocked and lock were both used (doesn't guarantee atomicity IMO).

### DIFF
--- a/prometheus-net/Counter.cs
+++ b/prometheus-net/Counter.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 using Prometheus.Advanced;
 using Prometheus.Advanced.DataContracts;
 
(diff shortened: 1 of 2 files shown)


andrasm:
Thanks Miha - merged, available in nuget v1.1.4

PR #23, opened 2016-07-14
Fix some metrics not updating, add process ID metric

qed-:


andrasm:
Thanks David, merged and available on nuget.

PR #25, opened 2016-08-05
Update the UTF-8 encoding to never use the byte order mark

sstorie:
This is an extension of https://github.com/andrasm/prometheus-net/pull/17, but was discovered while investigating https://github.com/andrasm/prometheus-net/pull/24. Even if PR24 isn't merged, the BOM should probably never be included in the encoded metrics.


sstorie:
Here's an SO comment that provides some context around the Byte Order Mark

http://stackoverflow.com/a/15583729/571237

sstorie:
I just realized I reference the push gateway in my comment, but that's not part of the code yet. If you want me to amend my comment in the code please let me know.

sderen:
you may want to initialize UTF8Encoding object only once instead of initializing every time.

sstorie:
Yes, great idea. I updated the PR.

PR #26, opened 2016-08-09
Threading improvements

Aberwyvern:
Updated the way threading is handled for Counter, Gauge, and Histogram so that locks are not required as Prometheus discourages the use of mutex pattern for client libraries. Adding some missing capabilities recommended by Prometheus.

### DIFF
--- a/prometheus-net.tests/MetricsTests.cs
+++ b/prometheus-net.tests/MetricsTests.cs
@@ -7,7 +7,8 @@ using Should;
 
 namespace Prometheus.Tests
 {
-    public class MetricsTests
+	[TestFixture]
+	public class MetricsTests
     {
         [SetUp]
         public void Setup()
@@ -33,6 +34,8 @@ namespace Prometheus.Tests
             var counter = Metrics.CreateCounter("name2", "help2", "label1");
             counter.Inc();
             counter.Inc(3.2);
+            Assert.Throws<InvalidOperationException>(() => counter.Inc(0));
+            Assert.Throws<InvalidOperationException>(() => counter.Inc(-1));
             counter.Value.ShouldEqual(4.2);
 
             counter.Labels("a").Value.ShouldEqual(0);
@@ -124,12 +127,10 @@ namespace Prometheus.Tests
             metrics[0].gauge.value.ShouldEqual(3.8);
         }
 
-      
-
         [Test]
         public void histogram_tests()
         {
-            Histogram histogram = Metrics.CreateHistogram("hist1", "help", new []{ 1.0, 2.0, 3.0});
+            Histogram histogram = Metrics.CreateHistogram("hist1", "help", new []{ 1.0, 2.0, 3.0, double.PositiveInfinity});
             histogram.Observe(1.5);
             histogram.Observe(2.5);
             histogram.Observe(1);
@@ -139,6 +140,7 @@ namespace Prometheus.Tests
             histogram.Observe(1.4);
             histogram.Observe(1.5);
             histogram.Observe(3.9);
+            histogram.Observe(double.NaN);
 
             var metric = histogram.Collect().metric[0];
             metric.histogram.ShouldNotBeNull();
@@ -149,6 +151,57 @@ namespace Prometheus.Tests
             metric.histogram.bucket[1].cumulative_count.ShouldEqual(5ul);
             metric.histogram.bucket[2].cumulative_count.ShouldEqual(8ul);
             metric.histogram.bucket[3].cumulative_count.ShouldEqual(9ul);
+        }
+
+        [Test]
+        public void histogram_default_buckets()
+        {
+            var histogram = Metrics.CreateHistogram("hist", "help");
+            histogram.Observe(0.03);
+
+            var metric = histogram.Collect().metric[0];
+            metric.histogram.ShouldNotBeNull();
+            metric.histogram.sample_count.ShouldEqual(1ul);
+            metric.histogram.sample_sum.ShouldEqual(0.03);
+            metric.histogram.bucket.Count.ShouldEqual(15);
+            metric.histogram.bucket[0].upper_bound.ShouldEqual(0.005);
+            metric.histogram.bucket[0].cumulative_count.ShouldEqual(0ul);
+            metric.histogram.bucket[1].upper_bound.ShouldEqual(0.01);
+            metric.histogram.bucket[1].cumulative_count.ShouldEqual(0ul);
+            metric.histogram.bucket[2].upper_bound.ShouldEqual(0.025);
+            metric.histogram.bucket[2].cumulative_count.ShouldEqual(0ul);
+            metric.histogram.bucket[3].upper_bound.ShouldEqual(0.05);
+            metric.histogram.bucket[3].cumulative_count.ShouldEqual(1ul);
+            metric.histogram.bucket[4].upper_bound.ShouldEqual(0.075);
+            metric.histogram.bucket[4].cumulative_count.ShouldEqual(1ul);
+        }
+
+        [Test]
+        public void histogram_no_buckets()
+        {
+            try
+            {
+                Metrics.CreateHistogram("hist", "help", new double[0]);
+                Assert.Fail("Expected an exception");
+            }
+            catch (ArgumentException ex)
+            {
+                Assert.AreEqual("Histogram must have at least one bucket", ex.Message);
+            }
+        }
+
+        [Test]
+        public void histogram_buckets_do_not_increase()
+        {
+            try
+            {
+                Metrics.CreateHistogram("hist", "help", new double[] { 0.5, 0.1 });
+                Assert.Fail("Expected an exception");
+            }
+            catch (ArgumentException ex)
+            {
+                Assert.AreEqual("Bucket values must be increasing", ex.Message);
+            }
         }
 
         [Test]
--- a/prometheus-net/Advanced/ThreadSafeDouble.cs
+++ b/prometheus-net/Advanced/ThreadSafeDouble.cs
@@ -1,0 +1,60 @@
+﻿using System;
+using System.Threading;
+
+namespace Prometheus.Advanced
+{
+    public struct ThreadSafeDouble
+    {
+        private long _value;
+
+        public ThreadSafeDouble(double value)
+        {
+            _value = BitConverter.DoubleToInt64Bits(value);
+        }
+
+        public double Value
+        {
+            get
+            {
+                return BitConverter.Int64BitsToDouble(Interlocked.Read(ref _value));
+            }
+            set
+            {
+                Interlocked.Exchange(ref _value, BitConverter.DoubleToInt64Bits(value));
+            }
+        }
+
+        public void Add(double increment)
+        {
+            while (true)
+            {
+                long initialValue = _value;
+                double computedValue = BitConverter.Int64BitsToDouble(initialValue) + increment;
+
+                //Compare exchange will only set the computed value if it is equal to the expected value
+                //It will always return the the value of _value prior to the exchange (whether it happens or not)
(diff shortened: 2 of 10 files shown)


prometheus-net.tests/MetricsTests.cs
```diff
@@ -124,12 +127,10 @@ public void gauge_collection()
             metrics[0].gauge.value.ShouldEqual(3.8);
         }
 
-      
-
         [Test]
         public void histogram_tests()
         {
-            Histogram histogram = Metrics.CreateHistogram("hist1", "help", new []{ 1.0, 2.0, 3.0});
+            Histogram histogram = Metrics.CreateHistogram("hist1", "help", new []{ 1.0, 2.0, 3.0, double.PositiveInfinity});
```
Aberwyvern: Added infinity here to test that infinity doesn't get added twice as there was no test for that. Other (new) test cases test without the infinity item to make sure it does get added as expected

prometheus-net/Advanced/ThreadSafeDouble.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using System.Threading;
+
+namespace Prometheus.Advanced
+{
+    public struct ThreadSafeDouble
+    {
+        private long _value;
```
Aberwyvern: Interlocked.Read works on longs.
Bit converter will allow us to store a double to a long with ease

prometheus-net/Counter.cs
```diff
@@ -26,37 +26,27 @@ public void Inc(double increment = 1)
...
             }
 
-            public void Inc(double increment = 1)
+            public void Inc(double increment = 1.0D)
             {
-                if (increment < 0)
-                {
-                    throw new InvalidOperationException("Counter cannot go down");
-                }
+                if (increment <= 0.0D)
```
Aberwyvern: According to prometheus documentation:
MUST check that v >= 0

prometheus-net/Counter.cs
```diff
@@ -26,37 +26,27 @@ public void Inc(double increment = 1)
...
+                    throw new InvalidOperationException("Counter can only go up");
 
-                double newCurrentValue = 0;
-                while (true)
```
Aberwyvern: This implementation was not quite right. 
It would cause double computations on everything exception the first increment.
The new logic is encapsulated in ThreadSafeDouble and follows MSDN recommendations
https://msdn.microsoft.com/en-us/library/cd0811yf(v=vs.110).aspx

prometheus-net/Counter.cs
```diff
@@ -26,37 +26,27 @@ public void Inc(double increment = 1)
...
-                        return;
-                }
+                _value.Add(increment);
             }
 
             public double Value
             {
                 get
                 {
-                    return Interlocked.CompareExchange(ref _value, 0, 0);
```
Aberwyvern: Unnecessary compare exchange.
The updated implementation in ThreadSafeDouble just does a read

prometheus-net/Gauge.cs
```diff
@@ -20,9 +20,26 @@ internal Gauge(string name, string help, string[] labelNames)
         }
 
 
+        public class Timer
```
Aberwyvern: Follows the spirit of the official Java client implementation

prometheus-net/Gauge.cs
```diff
@@ -32,22 +49,24 @@ protected override void Populate(Metric metric)
...
+                _value.Add(increment);
             }
 
             public void Set(double val)
             {
-                Interlocked.Exchange(ref _value, val);
+                _value.Value = val;
+            }
+
+            public void SetToCurrentTime()
```
Aberwyvern: Recommended part of the API as recommended by official documentation

prometheus-net/Gauge.cs
```diff
@@ -32,22 +49,24 @@ protected override void Populate(Metric metric)
...
+            {
+                var unixTicks = System.DateTime.UtcNow.Ticks - new System.DateTime(1970, 1, 1, 0, 0, 0, System.DateTimeKind.Utc).Ticks;
+                Set(unixTicks / System.TimeSpan.TicksPerSecond);
             }
 
+            public Gauge.Timer StartTimer()
```
Aberwyvern: Recommended part of the API as recommended by official documentation

prometheus-net/Histogram.cs
```diff
@@ -25,9 +24,9 @@ internal Histogram(string name, string help, string[] labelNames, double[] bucke
             }
             _buckets = buckets ?? DefaultBuckets;
 
-            if (_buckets.Length <= 1)
```
Aberwyvern: One bucket is acceptable.
Only zero buckets is unacceptable.

prometheus-net/Histogram.cs
```diff
@@ -48,55 +47,35 @@ internal Histogram(string name, string help, string[] labelNames, double[] bucke
 
         public class Child : Advanced.Child, IHistogram
         {
-            private double _sum = 0;
-            private ulong _count = 0;
-            private ulong[] _bucketCounts;
-            private readonly object _lock = new object();
-
-            //re-using WireFormat.Histogram to avoid excessive array allocations
```
Aberwyvern: Regarding these TODOs

1) Reusing the wire format is wonky because any duplicate request to obtain metrics will be operating off the same instance of the wireMetrics that we mod and return. This means that one request could be in the progress of emitting and rendering these metrics while another request is in the process of collecting and modding that same wireMetrics instance. The implementation was modified to produce a new wire metrics instance on collection call in a similar fashion to the official java client.  Was there any solid evidence that the allocation of arrays was causing a meaningful performance issue?

2) HdrHistogram is not a good candidate here because the library explicitly states that is is not thread-safe. It's possible that some of the optimizations they use could be incorporated here.  Perhaps that could be investigated?

prometheus-net/Histogram.cs
```diff
@@ -48,55 +47,35 @@ internal Histogram(string name, string help, string[] labelNames, double[] bucke
 
         public class Child : Advanced.Child, IHistogram
         {
-            private double _sum = 0;
-            private ulong _count = 0;
```
Aberwyvern: This can and should be computed during collection as you ideally want the total could on the result of the collection to match the count that makes up the buckets. If you have the count computed here like this then it can continue to update while other metrics are being collected if we remove locks (which we should as recommended by prometheus client library writing documentation)

prometheus-net/Histogram.cs
```diff
@@ -48,55 +47,35 @@ internal Histogram(string name, string help, string[] labelNames, double[] bucke
...
-            private readonly object _lock = new object();
-
-            //re-using WireFormat.Histogram to avoid excessive array allocations 
-            //TODO: this will break thread-safety if the very same histogram instance is registered on multiple CollectorRegistries
-            //TODO: we should investigate if HdrHistogram can be a good candidate to use here internally - that's heavily optimized
-
-            private Advanced.DataContracts.Histogram _wireMetric;
-            private double[] _buckets;
+            private ThreadSafeDouble _sum = new ThreadSafeDouble(0.0D);
+            private ThreadSafeLong[] _bucketCounts;
```
Aberwyvern: Even though we the data contract is using ulong here long will be sufficient.

The java client actually uses double for this count, but because of the precision on a double in java it will lose the ability to increment by 1 at a lower value than the long will. So technically, our long can "count" higher than the official java client can. That said, we could potentially rollover, so we may want to guard that, but we won't rollover until 9,223,372,036,854,775,808.  If we ran a program continuously for 100 years how many observations per second is that? 9,223,372,036,854,775,808 / (60 * 60 * 24 * 365 * 100) = 2.9 billion observations /second. I think we can safely assume that will not reasonably happen so I think we are safe to not worry about rollover here.

prometheus-net/Histogram.cs
```diff
@@ -48,55 +47,35 @@ internal Histogram(string name, string help, string[] labelNames, double[] bucke
...
+            private double[] _upperBounds;
             
             internal override void Init(ICollector parent, LabelValues labelValues)
             {
                 base.Init(parent, labelValues);
 
-                _buckets = ((Histogram) parent)._buckets;
-
-                if (_buckets == null)
```
Aberwyvern: This could never happen because the parent requires buckets to be defined.

prometheus-net/Histogram.cs
```diff
@@ -106,28 +85,15 @@ public void Observe(double val)
...
-                    if (val <= _buckets[i])
+                    if (val <= _upperBounds[i])
                     {
-                        bucketIndex = i;
+                        _bucketCounts[i].Add(1);
                         break;
                     }
                 }
-
-                if (bucketIndex == -1)
```
Aberwyvern: Infinity will always exist so this check is unnecessary

prometheus-net/Histogram.cs
```diff
@@ -106,28 +85,15 @@ public void Observe(double val)
...
-                {
-                    Trace.WriteLine(string.Format("Couldn't find a bucket for {0} in metric {1}", val, ToString()));
-                    return;
-                }
-
-                lock (_lock)
-                {
-                    ++_count;
```
Aberwyvern: See previous explanation about why count shouldn't be tracked here.
It can be computed as we traverse to the bucket counts to produce a more internally consistent representation of the count as we are producing the result of a particular collection/populate action.

prometheus-net/Histogram.cs
```diff
@@ -106,28 +85,15 @@ public void Observe(double val)
...
                     }
                 }
-
-                if (bucketIndex == -1)
-                {
-                    Trace.WriteLine(string.Format("Couldn't find a bucket for {0} in metric {1}", val, ToString()));
-                    return;
-                }
-
-                lock (_lock)
```
Aberwyvern: From Prometheus client library authoring guidance:
_In our experience the least performant is mutexes._
_Processor atomic instructions tend to be in the middle, and generally acceptable._
_Approaches that avoid different CPUs mutating the same bit of RAM work best, such as the DoubleAdder in Java’s simpleclient. There is a memory cost though._

By converting the bucket counts and sum to ThreadSafeDouble we can avoid the need to wrap this in a lock statement.

prometheus-net/Histogram.cs
```diff
@@ -106,28 +85,15 @@ public void Observe(double val)
                     return;
                 }
 
-                int bucketIndex = -1;
-                for (int i = 0; i < _buckets.Length; i++)
+                for (int i = 0; i < _upperBounds.Length; i++)
```
Aberwyvern: This aligns more closely with the implementation of the official java client

prometheus-net/Histogram.cs
```diff
@@ -48,55 +47,35 @@ internal Histogram(string name, string help, string[] labelNames, double[] bucke
...
-                    for (int i = 1; i < _bucketCounts.Length; i++)
+                    wireMetric.sample_count += (ulong)_bucketCounts[i].Value;
+                    wireMetric.bucket.Add(new Bucket
                     {
-                        _wireMetric.bucket[i].cumulative_count = _bucketCounts[i] + _wireMetric.bucket[i - 1].cumulative_count;
-                    }
-                    _wireMetric.sample_count = _count;
-                    _wireMetric.sample_sum = _sum;
+                        upper_bound = _upperBounds[i],
+                        cumulative_count = wireMetric.sample_count
```
Aberwyvern: The accumulated count for the current index is the same as the cumulative count for the current bucket. This avoids the need to lookup into a previous bucket which can cause differences in counts throughout due threading when we remove the lock.  By reading only one value per bucket we allow the total count and the counts of each bucket to remain internally consistent for a collection without using locks.

andrasm:
Prolific. Thanks for the improvements! These are now part of 1.2-rc on nuget.

PR #33, opened 2016-10-27
LabelValue escaped for ASCII-format

pavelhritonenko:
Hello.

I use pushgateway for pushing metrics and sometimes some of my workers stop pushing metrics. I see in trace output error 500 from push-gateway service. It seems like there is malformed request because of a requirement of escaping newlines, slashes and double quotes is not met in the current code.


andrasm:
Thanks for the PR. Merged and published to nuget.

PR #34, opened 2016-11-08
Flush metrics when stopping MetricsPusher

valentynpopov:
This is especially important for short jobs which finish before having a chance to push any metrics.


andrasm:
Many thanks. Merged & pushed to nuget.

PR #36, opened 2016-12-02
Introduce MetricsMiddleware for OWin & Netcore

stefc:
Here a not tested dotnet core aproach for implement the /metrics endpoint via a middleware (Owin). Not tested but perhaps a solution that works with ASP.MVC and NancyFx. 

I use the csproj Tooling of netcore


dustinchilson:
fyi @stefc  If you're interested in collaborating, 

https://github.com/dustinchilson/PrometheusNet.Owin

sandersaares:
Sounds like this PR is obsolete and outdated, so closing. ASP.NET Core middleware will be included in version 2.0.

PR #38, opened 2017-01-13
Added .NET 4.5 Support with Reactive 3.x - .NET 4.0 targeting Rx 2.2.5

lakario:
- Refactored main project to .shared project with .NET 4.5 and .NET 4 targets projects
- Refactored main test project to .sharedtests project with .NET 4.5 and .NET 4.0 targets projects (previously targeted just 4.5 for tests!)
- Updated nuspec definition to support multiple .NET targets.
  - NOTE: Due to this change, you must include version parameter when building nuget: `deploy_nuget.bat 1.3.4`
- Enabled documentation on build output
  - TODO: add documentation


lakario:
I just wanted to point out that the main reason I did this is that System.Reactive 3.x cannot be (easily) combined with Rx 2.x in a single assembly. I was attempting to instrument another library with System.Reactive 3.x and then funnel the data to prometheus-net, but because of prometheus-net's dependency on Rx 2.x, I was unable to combine the two implementations.

lakario:
Awesome, thanks!

andrasm:
Hi lakario - many thanks for the PR. Pushed to nuget as 1.3.4-beta and will remove the beta after a week or so prod usage.

andrasm:
yeah, I didn't spot this either - will revert it to 1.3 after the prod testing

sia-s:
Can you remove beta from 1.3.4? I've been using it in a test env for a month now without any problem. Our TeamCity does not allow pre-release nugets to be used for stable libraries.

PR #46, opened 2017-02-07
Tester metric server now uses less common port

sandersaares:
It used to use port 1234, which can conflict with other development systems since 1234 is just so easy to type.

Now it uses an arbitrary higher port number, hopefully making life easier for developers with other dev junk also installed and running on their PC.

### DIFF
--- a/tester/MetricServerTester.cs
+++ b/tester/MetricServerTester.cs
@@ -13,12 +13,12 @@ namespace tester
     {
         public override IMetricServer InitializeMetricHandler()
         {
-            return new MetricServer(hostname: "localhost", port: 1234);
+            return new MetricServer(hostname: "localhost", port: 21881);
         }
 
         public override void OnObservation()
         {
-            var httpRequest = (HttpWebRequest)WebRequest.Create("http://localhost:1234/metrics");
+            var httpRequest = (HttpWebRequest)WebRequest.Create("http://localhost:21881/metrics");
             httpRequest.Method = "GET";
 
             using (var httpResponse = (HttpWebResponse)httpRequest.GetResponse())


andrasm:
Thanks for the PR. Can you please change the port to something below the ephemeral range (typically below 32768) - binding to a port in that range is not ideal - although admittedly the chance of collision is small, and probably less bad than using the "common 1234".

sandersaares:
Okay, sure. Changed to port 21881.

judwhite:
@andrasm @sandersaares I have some C# code which picks a random available port (similar to Go's `httptest.NewServer`), would you be interested in this change?

@andrasm Unrelated to this PR, I'm going to start using this library and may knock out a few of the currently open issues. Are you available to review?

sandersaares:
Manually merged to v2 branch avoid merge conflicts.

@judwhite I am currently going through the open issues and PRs and hope to cover them all for a v2 release soon. Happy to review and receive PRs after v2 has settled and published - codebase might be too turbulent before that.

judwhite:
@sandersaares Thanks. I'm about to pick a Prom .NET library - would you suggest v1 or v2 at this point? I'm ok with an API in-flux if you think the code itself is reasonably reliable.

Here's the random port selector (there's probably a better way; this worked for me).

```cs
private static int GetRandomUnusedPort()
{
    // unfortunately HttpListener doesn't have this functionality
    var listener = new TcpListener(IPAddress.Loopback, 0);
    listener.Start();
    var port = ((IPEndPoint)listener.LocalEndpoint).Port;
    listener.Stop();
    return port;
}
```

This code needs to be in a retry loop because of occasional timing issues and no one likes a flaky test 😄

```cs
_httpListener = new HttpListener();
int port = GetRandomUnusedPort();
_url = string.Format("http://localhost:{0}/", port);
_httpListener.Prefixes.Add(_url);
if (authenticationSchemes != null) // optional
{
    _httpListener.AuthenticationSchemes = authenticationSchemes.Value;
}
_httpListener.Start();
```

sandersaares:
prometheus-net v1 works fine in terms of functionality - that's what I use in production right now. There are a few quirks here and there but no big problems (mostly described in open issues - which are generally all fixed in v2, just waiting for merge before I close them).

v2 has a few breaking changes in the interest of continued maintainability but the API surface itself is not very affected, just the internals are getting turned around and restructured. For users it should not change much - the changes just affect the code itself.

I expect to publish a preview package of v2 within a week and publish to stable perhaps in a month (assuming no surprises).

PR #48, opened 2017-03-16
added .NET 4.6 project to reference proper *.Reactive.* versions

MihaMarkic:
There is a version mismatch with *.Reactive.* assemblies when targeting .NET 4.6 (4.5 versions target 3.0.1000 while 4.6 target 3.0.3000).
So to avoid warnings I created a .NET 4.6 version (more or less copy&paste from 4.5).


andrasm:
Many thanks, Miha - merged and published to nuget as 1.3.5

lakario:
@MihaMarkic Thanks for fixing that!

PR #51, opened 2017-04-06
Optional timestamps on metrics

sandersaares:
You can now optionally specify a custom timestamp to attach to metrics, updated independently of the value.

This is intended mainly to enable export of historical metrics when bootstrapping a Prometheus database from an existing data set.

Did not add a SetTimestamp() on Summary, as I was thinking it might not make sense to report historical Summary data as Summary itself contains a set of time-related logic that might not work for historical data. Did not explore this in depth, so feel free to correct me if I am thinking along the wrong lines on this point.

### DIFF
--- a/prometheus-net/Advanced/Child.cs
+++ b/prometheus-net/Advanced/Child.cs
@@ -1,5 +1,7 @@
 using Prometheus.Advanced.DataContracts;
 using Prometheus.Internal;
+using System;
+using System.Threading;
 
 namespace Prometheus.Advanced
 {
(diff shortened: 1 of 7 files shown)


sandersaares:
PR updated to bring in latest commits from upstream.

kbalouse:
Can you clarify how this can be used to prefill Prometheus metrics?

Say I have a MetricServer with a counter with labels and I want to prefill historical data. Whats the correct way to do this?

I've gotten close in my attempts, but I'm seeing gaps in time series that are not updated often.

sandersaares:
The constraints on historical data backfilling are:

* Gaps in data are, in general, not tolerated by Prometheus or systems using Prometheus. 3 minutes between historical measurements seems a safe spot. At 5 minutes, data analytics started behaving wrongly.
* Prometheus needs timestamps in incrementing order, so you need to be careful about that.

How I solved it was by creating an exporter that starts from the beginning of the import moment. Then every time Prometheus queries it (needs experiment to avoid bottlenecks and timeouts; I think I used 30 seconds), the exporter provides a data dump and "moves time forward".

Now just by doing a normal export every time, the operation is quite slow - if Prometheus queries every 10 seconds and you do 3 minutes per cycle, that is still just at 18 times speed compared to real-time. The operation will take forever for larger batches!

However, Prometheus will happily ingest multiple values for the same series in the same import! So what I do is I do a large batch of data each time, with many 3-minute intervals per each Prometheus query. I believe I exported an entire day worth of data each time my exporter was queried. That way everything goes pretty fast (Prometheus disk write speed actually became a bit of a bottleneck).

@andrasm would it be possible to get this merged? It seems rather useful.

andrasm:
Can you please update the PR with the latest from master? You seemingly did this already but the .NET46 project files shouldn't be here - I might have screwed something when I merged the .net core support (PR 50) as that commit looks to be a no-op (might have rewritten history or something bad) - sorry if that's the case.

sandersaares:
PR updated with a clean rebase on master.

Note that I was not able to verify nothing broke as I do not know how to execute the tests anymore. Using NUnit.Console gives assmebly load errors (NUnit.Framework not found). I request instructions in the readme on what is the correct way to execute the tests project when using a clean install of Visual Studio with a fresh clone of the project.

brian-brazil:
Setting a timestamp is not an API that direct instrumentation should offer, that's solely the domain of custom collectors.

sandersaares:
I will close this for the time being, given that Prometheus 2.0 remote read features might be enough to fill this gap. If it continues to be of interest to anyone, please leave a comment.

PR #52, opened 2017-04-22
Add .Net Core 1.1 XPlatform support

ejkani:
Added .Net Core 1.1 Support. Since tests are not discoverable by Visual Studio 2017 yet, they have not been   executed, but they build OK.

MetricsServer and MetricsPusher are escaped with the compiler flag "NETCOREAPP1_1" so in this PR you need to implement  a "Metrics" endpoint sometning like this:

## startup.cs

    ...
    public static Lazy<DotNetStatsCollector> LazyDotNetStatsCollector = new Lazy<DotNetStatsCollector>(() => new DotNetStatsCollector());
    ...
    public void ConfigureServices(IServiceCollection services)
        {
    ...
            services.AddMvc(options =>
            {
                options.OutputFormatters.Add(new ProtobufOutputFormatter());
            });
    ...
        }
    ...
    public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
        {
    ...
            LazyDotNetStatsCollector.Value.RegisterMetrics();
    ...
        }

 ## MetricsController
        [HttpGet]
        public async Task<IActionResult> Get()
        {
            var registry = DefaultCollectorRegistry.Instance;
            var acceptHeaders = Request.Headers["Accept"];
            var contentType = ScrapeHandler.GetContentType(acceptHeaders);
            Response.ContentType = contentType;
            ObjectResult result;
            const string ProtoContentType = "application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited";

            Startup.LazyDotNetStatsCollector.Value.UpdateMetrics();
                using (var ms = new MemoryStream())
                {
                    ScrapeHandler.ProcessScrapeRequest(registry.CollectAll(), contentType, ms);
                    if (contentType == ProtoContentType)
                    {
                        if (ms.CanSeek) ms.Seek(0, SeekOrigin.Begin);
                        result = new ObjectResult(ms.ToArray());
                    }
                    else
                    {
                        var txt = Encoding.UTF8.GetString(ms.ToArray());
                        result = new OkObjectResult(txt);
                    }
                }
            return result;
        }

## ProtobufOutputFormatter 

    public class ProtobufOutputFormatter : OutputFormatter
    {
        static readonly MediaTypeHeaderValue ProtoMediaType = MediaTypeHeaderValue.Parse("application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited");
        private static readonly Lazy<RuntimeTypeModel> model = new Lazy<RuntimeTypeModel>(CreateTypeModel);

        public string ContentType { get; } = "application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited";

        public static RuntimeTypeModel Model => model.Value;

        public ProtobufOutputFormatter()
        {
            SupportedMediaTypes.Add(ProtoMediaType);
        }

        private static RuntimeTypeModel CreateTypeModel()
        {
            var typeModel = TypeModel.Create();
            typeModel.UseImplicitZeroDefaults = false;
            return typeModel;
        }

        public override Task WriteResponseBodyAsync(OutputFormatterWriteContext context)
        {
            var response = context.HttpContext.Response;
            return Task.FromResult(response);
        }
    }

### DIFF
--- a/prometheus-net.NET40.tests/packages.config
+++ b/prometheus-net.NET40.tests/packages.config
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="NUnit" version="3.0.1" targetFramework="net40" />
-  <package id="protobuf-net" version="2.0.0.668" targetFramework="net45" />
+  <package id="NUnit" version="3.6.1" targetFramework="net40" />
+  <package id="protobuf-net" version="2.1.0" targetFramework="net40" />
   <package id="Rx-Core" version="2.2.5" targetFramework="net40" />
   <package id="Rx-Interfaces" version="2.2.5" targetFramework="net40" />
   <package id="Should" version="1.1.20" targetFramework="net45" />
--- a/prometheus-net.NET40.tests/prometheus-net.NET40.tests.csproj
+++ b/prometheus-net.NET40.tests/prometheus-net.NET40.tests.csproj
@@ -31,13 +31,11 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="nunit.framework, Version=3.0.5813.39032, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
-      <HintPath>..\packages\NUnit.3.0.1\lib\net40\nunit.framework.dll</HintPath>
-      <Private>True</Private>
+    <Reference Include="nunit.framework, Version=3.6.1.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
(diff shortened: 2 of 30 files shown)


DianaD96:
Hi @ejkani ,
just a quick question...I checked your modified files and you seem to be using `PerformanceCounter` but this class is not available in .net core. Is there something I am missing? Thank you!

sunsided:
Any updates on this?

sandersaares:
Going forward from version 2.0 the library will support .NET Standard 2.0 and compatible runtimes, which does not include .NET Core 1.1. This is purely to enable low effort maintainability by one part-time developer, especially as the SDKs and toolchains before 2.0 were quite tricky to deal with.

Unfortunately this also means that any .NET Core 1.1 PRs cannot be merged. Therefore I am closing this PR.

PR #57, opened 2017-07-17
Fix ThreadSafeDouble.Add()

xakep139:
Since reading\writing of `long` is not thread safe (https://stackoverflow.com/a/5209632/4963622), reading of `_value` should be wrapped in `Interlocked.Read()` - https://msdn.microsoft.com/en-us/library/system.threading.interlocked.read(v=vs.110).aspx#Remarks


xakep139:
@andrasm it's up to you

sandersaares:
I believe @andrasm reasoning is correct here. Closing as not needed. If there does appear to be a problem, please provide code to demonstrate it.

PR #67, opened 2017-11-02
added support for netstandard2.0

alvillain:
Added support for netstandard2.0


sandersaares:
.NET Standard 2.0 support will be added in version 2.0 of the library. I will close this PR as unmerged since this feature is already present in the v2 branch.

PR #82, opened 2018-03-22
Expose all Summary defaults as public

discosultan:
This PR caters to a use case where the user would like to create a summary metric with mostly default parameters. Currently it's tricky because the more configurable overload requires specifying all the parameters. Even if the user figures out the current defaults, there's no way to indefinitely match them as this library may decide to change them in the future.

Since `Summary.DefObjectives` is already public, I propose also making the following defaults public:

- `Summary.DefMaxAge`
- `Summary.DefAgeBuckets`
- `Summary.DefBufCap`

It feels more cohesive if all the defaults are handled in a similar way. With that in mind, the values should also be marked `static readonly` instead of `const` so that they are always properly taken from this assembly at runtime.

### DIFF
--- a/Prometheus.NetStandard/Advanced/MetricFactory.cs
+++ b/Prometheus.NetStandard/Advanced/MetricFactory.cs
@@ -31,14 +31,7 @@ namespace Prometheus.Advanced
             return (Summary)_registry.GetOrAdd(metric);
         }
 
-        public Summary CreateSummary(
-            string name,
-            string help,
-            string[] labelNames,
-            IList<QuantileEpsilonPair> objectives = null,
-            TimeSpan? maxAge = null,
-            int? ageBuckets = null,
-            int? bufCap = null)
+        public Summary CreateSummary(string name, string help, string[] labelNames, IList<QuantileEpsilonPair> objectives, TimeSpan? maxAge, int? ageBuckets, int? bufCap)
         {
             var metric = new Summary(name, help, labelNames, objectives, maxAge, ageBuckets, bufCap);
             return (Summary)_registry.GetOrAdd(metric);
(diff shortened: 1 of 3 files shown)


discosultan:
I was actually thinking of the null pattern. I think that would be even better as this pattern is already in use and it would allow getting rid of the extra overload. Also, it removes the need to discover the static default params for this scenario.

I wasn't sure if the current multiple overload approach was deliberately made this way. Also, it's a slightly bigger change.

sandersaares:
I would not want to get rid of the overload - avoid a breaking change for users. But making the TimeSpan parameter nullable should be easy and backward compatible, while also keeping both overloads for any code that already uses them.

The reason for the multiple overloads is probably buried in deep history. If you wish to modify the PR to use instead the approach to make the maxAge parameter nullable, I will be happy merge it.

discosultan:
Oh right, I missed the usage of `params` keyword in the first overload. I agree - that wouldn't be a nice change.

I will update the second overload shortly.

sandersaares:
I think making them actual default parameters with `= null` might create ambiguity in overload resolution (what if there is just 1 labelName). It should be good enough if the maxAge parameter is just made nullable without actually making them default to null.

So you would do `CreateSummary(name, help, labelnames, null, null, ageBuckets: 5, null)` to use default values for those you do not care about specifying. This should preserve API compatibility in all existing usages as far as I can tell. Do you agree?

Yes, I think it is fine to leave the defaults public. I do not expect there to be any significant changes in the API surface of the Summary class going forward - it seems to work well as-is (and has been so for a few years already without changes being requested).

discosultan:
👍 Changes made.

sandersaares:

discosultan:
Thanks!

sandersaares:
2.1.0 with this PR is now published. For better future extensibility, I added an object-based configuration mechanism that allows the parameter set to be easily changed and selectively used while maintaining binary compatibility:

```
var hist = Metrics.CreateHistogram("my_histogram", "help text", new HistogramConfiguration
{
    Buckets = new[] { 0, 0.2, 0.4, 0.6, 0.8, 0.9 }
});
```

Perhaps this initialization style might also be useful to you. For future additions, this is where any new options would go, with the old style method overloads approach being for backward compatibility only.

PR #84, opened 2018-04-20
Modified invalid tests

omerfarukz:
Comment and assertion mismatched. Test seems wrong.


sandersaares:
In fact, it is the comment that is wrong - if there are no labels, the default behavior is to publish the metric immediately. Therefore we do expect to see data.

I will fix the comment - thank you for pointing it out!

PR #85, opened 2018-05-01
Create new CancellationTokenSource on each start

Patrickleijser:
Now you are able to restart the MetricServer.. else the CTS is null


sandersaares:
The MetricServer is not really designed to be restarted. Not saying it cannot be but is there a real world use case for this? Please explain the scenario of interest.

I would prefer to avoid having to keep all the different types of metric server restart-capable if the feature is not needed, just for simplicity of maintenance and implementation of the different classes. Right now the `if (_task != null)` check in MetricHandler should prevent restart if I read the code right (in addition to CTS being null).

sandersaares:
Closing this with timeout - if you wish to return to this, feel free to leave a comment.

PR #90, opened 2018-07-10
Replaces HTTP status code references with StatusCodes constants in Core to improve readability

dlowder:
Swaps out the HTTP status code references to make use of the StatusCodes constants available in Core here:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.statuscodes?view=aspnetcore-2.1


sandersaares:

PR #94, opened 2018-08-15
feat(collector) add queryParameters to collector

### LINKED ISSUE #93: Add query parameters to Collector interface
It would be nice to be able to receive query parameters in a custom collector. This way I can send arguments from prometheus to my custom built exporter (like it's done in the official [blackbox_exporter](https://github.com/prometheus/blackbox_exporter).

`http://my-custom-exporter/metrics?target=http:google.ch`

virth:
This adds potentially passed query parameters like `http://custom-exporter/metrics?target_url=http://google.com` to the ICollector interface to be considered by a custom collector.

- closes #93

### DIFF
--- a/Prometheus.AspNetCore/MetricServerMiddleware.cs
+++ b/Prometheus.AspNetCore/MetricServerMiddleware.cs
@@ -2,6 +2,7 @@
 using Prometheus.Advanced;
 using Prometheus.Advanced.DataContracts;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.IO;
 using System.Threading.Tasks;
 
@@ -9,7 +10,7 @@ namespace Prometheus
 {
     /// <summary>
     /// Prometheus metrics export middleware for ASP.NET Core.
-    /// 
+    ///
     /// You should use IApplicationBuilder.UseMetricServer extension method instead of using this class directly.
     /// </summary>
     public sealed class MetricServerMiddleware
@@ -50,7 +51,12 @@ namespace Prometheus
 
             try
             {
-                metrics = _registry.CollectAll();
+                var queryParameters = new NameValueCollection();
+                foreach (var query in request.Query)
+                {
+                    queryParameters.Add(query.Key, query.Value);
+                }
+                metrics = _registry.CollectAll(queryParameters);
             }
             catch (ScrapeFailedException ex)
             {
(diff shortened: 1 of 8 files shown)


virth:
I just squashed my commits into one. The PR is now ready for review.

@sandersaares could you, or somenone of the team, review my changes?

sandersaares:
I would like to learn more about the types of collectors that would benefit from this functionality. I feel a bit uneasy about adding such a generic input mechanism - would it really be used except for very custom things? Prometheus exporters seem to generally be far more passively designed, with blackbox being something of an exception. What exactly is your scenario?

PS. Adding a parameter to a method, even with a default value, is a runtime breaking change because default parameters are compile-time resolved. This change would need to add a new overload to the interface to avoid breaking existing deployments with custom collectors.

virth:
Thanks for having a look. 

### why

What I actually wanna accomplish is an exporter, that is capable of calling any HTTP endpoint and gather some metrics of that endpoint. (Like the blackbox exporter, but different metrics)

With this library I'm able to accomplish this. But since I can not pass any argument to my custom exporter, I have to configure the list of endpoints directly inside my custom exporter. 

With the possibility of passing arguments to my exporter I could configure the list of endpoints inside prometheus. A call from prometheus to my exporter would then look somethings like this `http://my-exporter/metrics?target=https://google.com`. This is literally what the blackbox exporter does, with the major difference that I wanna collect different metrics than the blackbox exporter.

With the actual goal of configuring everything in prometheus I would have two major benefits:

1. The configuration, which targets to scrape, sites right beneath prometheus. Where all other configurations for other exporters sit anyway.
2. I can use prometheus mechanisms like relabelling and scraping individual targets at individual intervals. (Which I would have to implement in my exporter myself or maybe even run multiple instances of my exporter which can be scraped at different times)

### breaking change

As for the mentioned breaking change. I can change this so it's not breaking. But I would be happy to now first, if you even consider merging it 😄 

Thanks in advance!

sandersaares:
Thank you for the explanation. I think we need more opinions here from experienced Prometheus users and developers to be able to determine whether this is something that a Prometheus client library should implement. I can see arguments on both sides, so I try to be cautious here.

I would ask that you post in the [prometheus-users group](https://groups.google.com/forum/#!forum/prometheus-users) and ask for opinions there on whether this addition to the library is a good way to achieve your goals. The main group of Prometheus developers frequent that list and if they think this is a good idea, I would be much more comfortable merging this.

virth:
Sounds like a plan. I'll link my post here.

sandersaares:
I see that Brian commented on the issue and gave some explanation on why this would not be a good way to go. I will abide by his opinion here and thus close the PR as rejected.

I think prometheus-net should be sufficiently flexible/extensible to support a custom registry scenario as Brian described (I know I have successfully done such things in the past) but if there seems to be a blocker, please feel free to open a new issue or PR.

Issue #96, opened 2018-09-13
Change metric type names to lowercase


danielfm:
After upgrading to Prometheus v2.4.0, the metrics endpoint exposed by my .NET services stopped working with the following error: `invalid metric type "HISTOGRAM"`

![image](https://user-images.githubusercontent.com/29534/45493546-5d417f00-b745-11e8-971d-b2eca30790db.png)

A related issue has been opened in the Prometheus repo, but apparently the fix should be done in the client library, as it's not reporting the metric names as documented [here](https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md).

Ref: https://github.com/prometheus/prometheus/issues/4602

**Edit.** More information:

- Client version 1.3.5
- .NET version: 4.5.2

sandersaares:
The fix will be published as 2.1.3 pre-release package within the next 60 minutes. Please confirm that you no longer experience the issue with this version. I will publish the fix as a stable version once I hear back from you.

rgl:
I confirm it works here :-)

sandersaares:
Thanks for the confirmation. I have pushed the button to publish as stable. It should show up when nuget.org does their refresh.

rgl:
can you also do a github release?

PR #97, opened 2018-09-14
convert the metric family type to lowercase

rgl:
this should fix https://github.com/prometheus-net/prometheus-net/issues/96


inaciocorrea:
up

sandersaares:
Looks good. Thanks for the contribution!

PR #110, opened 2018-12-07
Allow multiple metrics with the same name but with the different labels

ABaboshin:
Goal of this PR is to have metrics like this:

```
my_app{"dependency=myservice1", "status=ok"}
my_app{"dependency=myservice2", "status=ok"}
my_app{"dependency=myservice3", "status=failed"}
```

With the current implementation it's not possible. 
So I changed the way how the keys for the collectors are going to be built.

### DIFF
--- a/Prometheus.NetStandard/Advanced/DefaultCollectorRegistry.cs
+++ b/Prometheus.NetStandard/Advanced/DefaultCollectorRegistry.cs
@@ -84,10 +84,6 @@ namespace Prometheus.Advanced
         {
             var key = $"{collector.Name}|{string.Join("|", collector.LabelNames ?? new string[] { })}";
             var collectorToUse = _collectors.GetOrAdd(key, collector);
-
-            if (!collector.LabelNames.SequenceEqual(collectorToUse.LabelNames))
-                throw new InvalidOperationException("Collector with same name must have same label names");
-
             return collectorToUse;
         }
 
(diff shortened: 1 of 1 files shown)


SukharevAndrey:
Very useful change. Library author, please merge this.

sandersaares:
The text snippet posted is not a valid Prometheus export document.

The code allows metrics to be exported with different label names. This is not consistent with existing Prometheus implementations and therefore unlikely to be interoperable with existing software. Therefore I cannot merge the PR, as it would not work well in the ecosystem.

Furthermore, I do not exactly understand the use case. You should use the same set of label names for a metric, although you may export the same metric with different values.

Example of valid metrics export with different label values but the same label names:

```
my_app{dependency="myservice1",status="ok"}
my_app{dependency="myservice2",status="ok"}
my_app{dependency="myservice3",status="failed"}
```

If you truly think this is the right way to do things, please post in https://groups.google.com/forum/#!forum/prometheus-users and describe your scenario and proposed solution in detail. If your solution receives positive feedback in that forum, I could be convinced to change my mind on this.

ABaboshin:
Ok, I managed to do what I want without changing the library in the following way:

```c#

// create i.e. a histogram

var hist = Metrics.CreateHistogram("my_app", "", new Prometheus.HistogramConfiguration
                {
                    LabelNames = new[] { "dependency", "status" }
                });

// update it later with your values

hist.WithLabels("dep1", "ok");
hist.Publish();

// and later again

hist.WithLabels("dep2", "failed");
hist.Publish();

```

As a result you will have 

```
my_app{dependency="dep1",status="ok"}
my_app{dependency="dep2",status="failed"}
```

PR #111, opened 2018-12-31
Add Exponential and Linear histogram buckets functions

cmannix:
Useful helper functions for defining a set of buckets for histograms. Taken directly from the Go implementation [here](https://github.com/prometheus/client_golang/blob/master/prometheus/histogram.go)


sandersaares:
Looks good but please inline the `CheckArgument` calls as that is not very idiomatic C# and this style of argument checking is not used elsewhere in the project.

cmannix:
Thanks for taking a look - done 👍

Issue #83, opened 2018-04-12
Built-in HTTP server metrics middleware


ozkar99:
any plans on adding http_server metrics on the middleware?

talking about duration of requests or total count of requests, like the ruby or go client does.

sandersaares:
There are currently no such plans. Would you be interested in contributing such a feature? If so, we could discuss details about how and what to implement.

ozkar99:
Definetly,

i've noticed that:  https://github.com/PrometheusClientNet/Prometheus.Client.HttpRequestDurations
already implements the functionality, but since this is THE dotnet ofically unoficial (since its linked on the prometheus clients page) 

I could start by trying to port that functionality.

sandersaares:
Taking reference from the [official Go implementation](https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/instrument_server.go) would be ideal, as it seems to exemplify a good number of Prometheus best practices.

With any web request instrumentation, it would be important to avoid an explosion in the number of exported metrics simply because many different URLs are accessed by users. The Go implementation only applies Code and Method labels by default, which I think is a good choice and helps keep this under control.

It would be nice to have an implementation that is natively focused on ASP.NET Core concepts like controllers and actions rather than path-string parsing, as in the linked implementation. Perhaps the implementation could auto-detect some "controller" and "action" labels instead of focusing on paths? I see there is some relevant info in a [ControllerActionDescriptor class](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.controllers.controlleractiondescriptor?view=aspnetcore-2.0), though I do not know from where the middleware would get an appropriate instance of this class.

Naturally, not every HTTP request goes to a controller/action. Such requests would have these labels with empty values.

Go implementation exposes the following metrics:

* requests currently being processed
* duration (histogram/summary)
* request count
* time until response headers are written
* request size (histogram/summary)
* response size (histogram/summary)

That is a good set of metrics to target, in general. If I read the code right, it is up to the user of the code to decide if they want a histogram or summary. That's rather nice in terms of flexibility, though I suspect prometheus-net does not currently support it (but it might be easy to add this abstraction).

An implementation along these lines would be very welcome. I would be happy to merge one with only partial metrics from the Go list (whatever is easiest to implement). The crucial point would be how to arrange the integration in a way that series explosion due to many different labels would be avoided.

[...]

PR #112, opened 2018-12-31
Add built in ASP.net HTTP request metrics

cmannix:
Based on the discussion here #83, this is an implementation of some middleware to export:

* Number of in progress  requests
* Request duration
* Request count.

This is still in progress, and needs a bit of work to make it ready, especially with regards to testing but as I may not be able to come back to this for a few days I would welcome input on the general approach of the work done so far.


Prometheus.HttpExporter.AspNetCore/HttpRequestCount/HttpRequestCountOptions.cs
```diff
@@ -0,0 +1,11 @@
+namespace Prometheus.HttpExporter.AspNetCore.HttpRequestCount
+{
+    public class HttpRequestCountOptions : HttpExporterOptionsBase
+    {
+        public Counter Counter { get; set; } =
+            Metrics.CreateCounter(DefaultName, DefaultHelp, "code", "method", "action", "controller");
+        
+        private const string DefaultName = "aspnet_http_request_count";
```
sandersaares: The standard style is to name "total count" counters with `_total` suffix, so `aspnet_http_requests_total` would be better here.

Prometheus.HttpExporter.AspNetCore/HttpRequestDuration/HttpRequestDurationOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Prometheus.HttpExporter.AspNetCore.HttpRequestDuration
+{
+    public class HttpRequestDurationOptions : HttpExporterOptionsBase
+    {
+        public Histogram Histogram { get; set; } = Metrics.CreateHistogram(DefaultName, DefaultHelp, Histogram.ExponentialBuckets(0.0001, 1.5, 36), "code", "method", "action", "controller");
+
+        private const string DefaultName = "aspnet_http_request_duration";
+
+        private const string DefaultHelp =
+            "Provides the duration in milliseconds of HTTP requests from an ASP.NET application.";
```
sandersaares: Prometheus standard style is to measure time in seconds, not milliseconds.

Prometheus.HttpExporter.AspNetCore/HttpRequestMiddlewareBase.cs
```diff
@@ -0,0 +1,78 @@
...
+
+                this.UpdateMetricValueIfExists(LabelNames.Method, context.Request.Method);
+                this.UpdateMetricValueIfExists(LabelNames.Code, context.Response.StatusCode.ToString());
+                this.UpdateMetricValueIfExists(LabelNames.Action, routeData?.Values["Action"] as string ?? string.Empty);
+                this.UpdateMetricValueIfExists(LabelNames.Controller, routeData?.Values["Controller"] as string ?? string.Empty);
+            }
+            else
+            {
+                this.UpdateMetricValueIfExists(LabelNames.Method, context.Request.Method);
+                this.UpdateMetricValueIfExists(LabelNames.Code, context.Response.StatusCode.ToString());
```
sandersaares: These two lines appear to be duplicated on both sides of the if/else. Unless I miss something obvious, it seems that they can be moved outside and de-duplicated.

Prometheus.HttpExporter.AspNetCore/HttpRequestMiddlewareBase.cs
```diff
@@ -0,0 +1,78 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Prometheus.Advanced;
+
+namespace Prometheus.HttpExporter.AspNetCore
+{
+    public class HttpRequestMiddlewareBase<T> where T : ICollector
```
sandersaares: It would be nice to see some documentation comments on this class, to outline the basic operating princioples of the label-name-based logic that is implemented. As-is it requires reading the code to discover that the choice of executed logic is driven by labels with specific names.

Prometheus.HttpExporter.AspNetCore/InFlight/HttpInFlightOptions.cs
```diff
@@ -0,0 +1,11 @@
+namespace Prometheus.HttpExporter.AspNetCore.InFlight
+{
+    public class HttpInFlightOptions : HttpExporterOptionsBase
+    {
+        public Gauge Gauge { get; set; } =
+            Metrics.CreateGauge(DefaultName, DefaultHelp);
+        
+        private const string DefaultName = "aspnet_http_inflight";
```
sandersaares: Standard naming style is to have the last word describe the units being measured. In this case the units are just individual requests, so better name would be `aspnet_http_inflight_requests`.

I am unsure about the term "inflight" here. Possibly "executing" would be more common terminology when considering readers coming from different platforms.

Prometheus.HttpExporter.AspNetCore/HttpRequestDuration/HttpRequestDurationOptions.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Prometheus.HttpExporter.AspNetCore.HttpRequestDuration
+{
+    public class HttpRequestDurationOptions : HttpExporterOptionsBase
+    {
+        public Histogram Histogram { get; set; } = Metrics.CreateHistogram(DefaultName, DefaultHelp, Histogram.ExponentialBuckets(0.0001, 1.5, 36), "code", "method", "action", "controller");
+
+        private const string DefaultName = "aspnet_http_request_duration";
```
sandersaares: Standard metric naming style puts the units last, so  `aspnet_http_request_duration_seconds` would be better.

Prometheus.HttpExporter.AspNetCore/HttpRequestDuration/HttpRequestDurationMiddleware.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            var stopWatch = new Stopwatch();
+            
+            stopWatch.Start();
+            await this.next(context);
+            stopWatch.Stop();
+
+            var labelData = GetLabelData(context);
+            
+            if (labelData != null) {
```
sandersaares: It is unclear to me what the purpose of the if statement is. As far as I can tell, GetLabelData() never returns null.

Prometheus.HttpExporter.AspNetCore/HttpRequestDuration/HttpRequestDurationMiddleware.cs
```diff
@@ -0,0 +1,38 @@
...
+            this.next = next ?? throw new ArgumentNullException(nameof(next));
+            this.requestDuration = histogram;
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            var stopWatch = new Stopwatch();
+            
+            stopWatch.Start();
+            await this.next(context);
```
sandersaares: An exception from `next()` would make the duration not be measured here. I think we should measure duration even in case of an exception here.

Prometheus.HttpExporter.AspNetCore/HttpRequestDuration/HttpRequestDurationMiddleware.cs
```diff
@@ -0,0 +1,38 @@
...
+        public HttpRequestDurationMiddleware(RequestDelegate next, Histogram histogram)
+         : base(histogram)
+        {
+            this.next = next ?? throw new ArgumentNullException(nameof(next));
+            this.requestDuration = histogram;
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            var stopWatch = new Stopwatch();
```
sandersaares: There's a Stopwatch.StartNew() that can condense these two lines together.

cmannix:
Hi - I've added some more tests and some documentation - I think this is pretty much there. Just a thought, having seen your most recent commit adding a link to the DotNetRuntime, do you think that this might be better off separate to the main project too?

sandersaares:
Thanks! I will do another review round when I get a moment.

I think no need to split into a separate repo at the moment. I will publish it as a separate NuGet package to keep dependencies isolated but the code can be in one bulk to keep life simple for everyone. After all, nobody wants to wonder if they are reporting issues in the right repo.

cmannix:
Thanks - I've just added some benchmarks on my branch and plan to do a small rename of the new projects to better reflect the frameworks used - will update this tomorrow with the changes.

sandersaares:
BenchmarkDotNet is often a source of many problems in how it interacts with the .NET SDK. The different "job types" to target both .NET Core and .NET Framework historically did not really work right.

Actually in my local clone I had already removed the .NET Framework benchmark project entirely and replaced it with a .NET Core runner project to keep things simple. It was only .NET Framework oriented in prometheus-net because when I added it the .NET Core runner did not work right. Today, the situation seems reversed - latest BenchmarkDotNet seems to work better on/with .NET Core than on .NET Framework!

sandersaares:
Thanks for all the effort! I will now merge this. Noticed a few small nits but I'll take care of them so there is no need for more back and forth. My recent push also made a few conflicts but no worries - I will resolve it during merge.

sandersaares:
Version 3.0.0 is now published, with this change included.

cmannix:
Thanks for all your help with this, I appreciate it 👍

PR #121, opened 2019-02-03
add sample size option for building histogram;

zmey-gorynych:
Avoid distribution bias by providing (currently hard-coded to 1L) sample size in Observe method of the Histogram. 

BACKGROUND:

In high frequency or batch processing scenarios it is incorrect to account using batch level metric when batches are variable size. In high frequency scenario the cost of accounting (for-each loop instead of binary search etc) may surpass the cost of measured event by 1-2 order(s) of magnitude rending instrumentation impractical.

The simplest solution is to aggregate random sample and record it using average sample measurement and sample size. That is what this change provides.

The Summary is more complicated as it needs many more changes in buffering algorithm. It would probably be best to provide new type of Summary that correctly accounts for sample size.

### DIFF
--- a/Prometheus.NetStandard/Histogram.cs
+++ b/Prometheus.NetStandard/Histogram.cs
@@ -6,6 +6,12 @@ namespace Prometheus
 {
     public interface IHistogram : IObserver
     {
+        /// <summary>
+        /// observe event supporting high frequency or batch processing use cases utilizing pre-aggregation
+        /// </summary>
+        /// <param name="val">average sample value</param>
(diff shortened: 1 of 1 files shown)


sandersaares:

zmey-gorynych:
You are welcome! Let me know when do you plan to release new update :) Cheers

sandersaares:
I expect to publish the new update this week, possibly even today. Doing some final validation with my own projects right now to ensure that nothing breaks with the new version.

sandersaares:
Version 3.0.0 is now published, with this change included.

PR #128, opened 2019-02-08
IObserver duration extensions

sanych-sun:
Convenient way to observe a duration (fluent syntaxes, supports async)

### DIFF
--- a/Prometheus.NetStandard/ObserverExtensions.cs
+++ b/Prometheus.NetStandard/ObserverExtensions.cs
@@ -1,0 +1,61 @@
+﻿using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+namespace Prometheus
+{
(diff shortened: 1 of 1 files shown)


sandersaares:
Is this not just `IObserver.NewTimer()` with a different syntax? (Wrapper method instead of using block)

sanych-sun:
Basically yes, but proposed syntax looks more intuitive for me. It's obvious that time would be measured over the provided action. In case of the timer it could be surprise that there is implicit observation on dispose. And avoid pitfall with async like this:

```
using(var timer = observer.NewTimer())
{
    return DoAsync();
}
```

I mean there is await inside the extension, so client code don't have to await.

Thanks.

sandersaares:
I can see some merit to that but I am going to hold off on this for now as I prefer not having two ways to do the same thing in the API just for providing a different API surface style.

PR #133, opened 2019-02-19
Dispose of MemoryStream in MetricPusher

### LINKED ISSUE #132: Undisposed stream in MetricPusher
A static analyzer is complaining about an undisposed MemoryStream in the prometheus-net package. I've tracked it down to here:

https://github.com/prometheus-net/prometheus-net/blob/master/Prometheus.NetStandard/MetricPusher.cs#L75

Could you please wrap this thing in a `using` statement?

VictorGavrish:
Fixes #132


sandersaares:

PR #135, opened 2019-02-24
don't allocate Collector objects unless it's needed

kimbell:
This is my proposal for reducing the number of objects being created. Some logic has been moved from MetricsFactory to CollectorRegistry; these two could be merged together. 

CollectorRegistry is the location where the ConcurrentDictionary is located. The idea is to only allocate the Collector objects if they are not present in the dictionary.


sandersaares:
Will solve this in a slightly different way to avoid spilling the initialization code too much into the registry.

PR #149, opened 2019-06-20
Change to use DateTime.UtcNow when working out the Summary Expiry Time

Brumawen:


sandersaares:
Thank you for the contribution! It is not immediately clear to me why this change is needed. Please explain the rationale.

Brumawen:
The _headStreamExpTime and _hotBufExpTime are currently calculated based on the local time (which is UTC+2 where I am).  The check to see if the buffers need to be flushed is done using the current utc time.  This means that the buffers will not be flushed the first time until 2 hours + _maxAge.   The change is to fix this.

sandersaares:
Aha, I see what you mean. Good catch!

sandersaares:
3.1.4 with this fix has been published and will be available on nuget.org as soon as the relevant caches update.

Brumawen:
Awesome.  Thanks so much for your very quick response.

PR #151, opened 2019-07-19
Extract prometheus-net.AspNetCore.Kestrel package

lodejard:
Layers Kestrel feature into additional Prometheus.AspNetCore.Kestrel package. This helps reduce dependency footprint when only the middleware is needed.

Also narrows Prometheus.AspNetCore dependencies to abstractions instead of implementations. This helps the core package work better on a wider range of web apps.

For example, a web app or worker process might depends on the bare minimum needed to host an AspNetCore endpoint. If the Prometheus.AspNetCore middleware package depends only on abstractions then it won't cause the rest of the Microsoft.AspNetCore framework to be added as a side-effect.


lodejard:
Not tested on full framework yet. I'm on a Mac so far this morning.

sandersaares:
This seems a potentially valuable change but I would like to understand it in more detail. What are some specific examples of scenarios where it has relevance, of apps that do not wish to pull in the entire AspNetCore but still want part of it? What is the practical benefit beyond dependency management (does it give a smaller footprint in some measures etc)? What is the impact on your specific project?

lodejard:
Hi, yes there are a number of reasons that as a general principle it is best for components packages to depend on a reduced set of abstraction packages, while applications depend on the complete implementation packages.

One is that it is really the application author who should control the policy of which version of the aspnetcore framework is bundled with their deployment. If they add a reference to a package which contains middleware, ideally that reference won't also upgrade the web application's version of aspnetcore as a side-effect.

A related example is when an application depends on components from several different nuget packages. The smaller the component's dependency graphs are, the less likely they are to introduce overlaps for the package manager to reconcile.

In my case, there is some utility functionality I was thinking of packaging to reuse. It's too vague to use as a concrete example, I'm not sure in practice if it would depend on the prom-net.NetStandard or prom-net.AspNetCore layer. But if it does end up depending on the prom-net.AspNetCore layer then this PR would enable packages like that to also have narrower abstractions-only impact on applications when they are added.

That said - those are dependency management cases. The practical footprint example is that it is possible to host an AspNetCore server without depending on the outermost "Microsoft.AspNetCore" package that brings in the entire system. (That package didn't even exist in the initial release, it was added later for convenience.) In that case the resulting build output could have a much smaller number of assemblies to deploy.

lodejard:
Any more thoughts on this? Let me know if you'd like it structured differently and I'd be happy to make adjustments.

In the meantime I've duplicated the middleware logic in our in-house infrastructure -- that way our infrastructure packages won't affect the application-layer dependency on aspnetcore.

sandersaares:
I did some looking into this and this does not seem like something I would want to pull at the moment. In general, the ASP.NET Core component of prometheus-net is not designed to be used wihout the full ASP.NET Core installed.

While I can see some theoretical cases where this might be an issue, the added maintenance burden in tracking all of these dependencies and keeping the numbers straight does not appear to be worth the benefit (which seems highly theoretical to me). From a maintenance point of view, I just want to declare "this library targets ASP.NET Core 2.1" and be done with it. Even this is made difficult enough by Microsoft (e.g. Microsoft.AspNetCore package is stuck on 2.1.7 but Microsoft.AspNetCore.All/App are 2.1.13 but officially do not support .NET Framework so what should I do for my .NET Framework users...).

Furthermore, I notice that ASP.NET Core 3.0 release notes say that some (many? all?) of the abstractions libraries are no longer produced, which makes me think "bullet dodged" but I admit I did not look into why they changed and what they changed there in terms of the abstractions.

PR #160, opened 2019-10-10
Allowing specifying a port to host metrics

rwkarg:
For example, this allows metrics to be served only over port 5001. This way API endpoints can be exposed over 5000 (or any other port) without also exposing potentially internal metrics.

I looked at updating the existing extension method but simply adding an `int? port` at the end of the method (to avoid a breaking change by reordering parameters) but:
- the ordering of parameters with path at beginning and port at end was odd
- shoving all implementation in one method was a bit complicated to follow

I'm open to opinions on placing `PathAndPort` in a separate file if that better matches the desired repository style.

### DIFF
--- a/Prometheus.AspNetCore/MetricServerMiddlewareExtensions.cs
+++ b/Prometheus.AspNetCore/MetricServerMiddlewareExtensions.cs
@@ -41,28 +41,27 @@ namespace Prometheus
                 Registry = registry
             });
         }
+    }
 
-        public class PathAndPort
+    public class PathAndPort
+    {
+        public PathAndPort(string path, int port)
         {
-            public PathAndPort(string path, int port)
-            {
-                Path = path;
-                Port = port;
-            }
-
-            /// <summary>
-            /// Creates a <see cref="PathAndPort"/> with the default URL <c>/metrics</c>, which is a Prometheus convention.
-            /// </summary>
-            /// <param name="port"></param>
-            /// <returns></returns>
-            public static PathAndPort WithDefaultPath(int port)
-            {
-                return new PathAndPort("/metrics", port);
-            }
-
-            public string Path { get; }
-            public int Port { get; }
+            Path = path;
+            Port = port;
         }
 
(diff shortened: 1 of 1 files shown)


sandersaares:
You do not need to modify the prometheus-net helper functions in order to add port limitations like that. If you want to filter the conditions on which the metric server is exported, you can just do the port limitation in your own app's startup code:

```csharp
// Specify URL prefix before filter, then no URL prefix later to avoid double prefix.
app.Map("/metrics", app2 => app2.MapWhen(c => c.Connection.LocalPort == 44327, app2 => app2.UseMetricServer("")));
```

rwkarg:
I could also register the middleware directly without using the existing extension method but it has been included as a convenience. I was adding functionality that matches the ASP.NET Core health check's UseHealthCheck extension method that allows both a path or a path and port to be specified.

I already have the extension method being used locally.

Is there a specific reason for the including the existing convenience extension method but not one that allows a port to be specified?

sandersaares:
There is no specific reason, no - but the intent is to keep it minimal. However, for compatibility with equivalent extensions in other libraries, I can see it might have value. I will reopen for another round of consideration later this week.

rwkarg:
A potentially cleaner method signature from a usage standpoint is to have the new extension method take an initial, required `int port` and then have the same optional parameters as the existing extension method.

That way, usage is simple:
`app.UseMetricServer(5001);`

It's not 100% ideal if a path is specified (naturally I'd expect to set the path first, then port) but it's not the end of the world:
`app.UseMetricServer(5001, "/metricz");`

sandersaares:
I suppose no approach is perfect here. ASP.NET Core endpoint routing has managed to improve on matters by having `.RequireHost()` but that's neither here nor there since prometheus-net is compatbile also with non-endpoint-routing and targets older versions of ASP.NET Core, as well.

Thank you for the contribution! I will merge this in its latest revision - hopefully it makes life easier for some users and it is small enough not to be a maintenance problem.

PR #166, opened 2019-10-28
Support for OData

lilbaek:
When using this library together with OData (https://github.com/OData/WebApi) the controller label remains empty. The OData team has decided to store the controller name using the key "odataPath" instead of "Controller". 
This PR adds support for writing the OData controller name to the controller label.

Before:
image

After:
image


sandersaares:
Thank you for the contribution, this looks good! I will also try to add a few words into the readme - perhaps you can review my text to verify that I understand the functionality correctly.

sandersaares:
I see this mechanism keeps the parameter names in the controller name (`ItemKey`), which does not happen with the regular ASP.NET Core controllers that we track. I am not strictly opposed to this but as one who is not familiar with OData in much depth, I would like to hear what the rationale for this design decision is.

sandersaares:
Hmmm I see in the controller string extraction a little possible issue.

Wouldn't this happen? `CopyFromItem(Name='Item')` -> `CopyFrom(Name=)` 

Just based on reading the code, there's nothing in there that constrains the replacement to the right location. This might result in unexpected results if you have short values like just "A" and whatnot. Or am I missing something in this logic?

Issue #164, opened 2019-10-26
ASP.NET package targetting netcoreapp3.0


kerams:
Hey, would you be OK with creating a netcoreapp3.0 target for Prometheus.AspNetCore? Right now, when I reference the package in a ASP.NET core 3 project, the Microsoft.AspNetCore dependency brings a lot of baggage that pollutes the lock file, even though it's redundant for that particular target. I could send in a PR, it should only be a couple of lines.

sandersaares:
Sure thing, will be published in next version. Thanks @kerams and @JosephWoodward for taking action on this!

giggio:
Was this ever fixed? I just noticed that 3.0 or 3.1 are still not there.

![image](https://user-images.githubusercontent.com/334958/85580954-8e392e00-b612-11ea-98cf-1207d9bb3977.png)

giggio:
Nevermind, I was not looking at the ASPNET Nuget. It is there. Sorry about that. https://www.nuget.org/packages/prometheus-net.AspNetCore

PR #174, opened 2019-11-29
Add netcoreapp3.0 as a target

JosephWoodward:
As ASP.NET core app is part of the shared framework there's no need to reference the package if targeting `netcoreapp3.0` TFM.

This addresses the issue raised in #164

### DIFF
--- a/Prometheus.AspNetCore/Prometheus.AspNetCore.csproj
+++ b/Prometheus.AspNetCore/Prometheus.AspNetCore.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
     <RootNamespace>Prometheus</RootNamespace>
 
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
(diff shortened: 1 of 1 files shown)


blankensteiner:
I'm wondering if it could also solve the problem if we just only depend on Microsoft.AspNetCore.Http.Abstractions and Microsoft.AspNetCore.Routing.Abstractions.
I don't know if we should be depending on meta-packages? Any thoughts in regards to this?
This leaves us with the problem of KestrelMetricServer, but as it noted "The overall utility of this class is somewhat suspect, considering HttpListener is available everywhere", so maybe it can be removed and moved to another NuGet package?

kerams:
@blankensteiner https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-3.0&tabs=visual-studio#remove-obsolete-package-references
>A large number of NuGet packages aren't produced for ASP.NET Core 3.0. Such package references should be removed from your project file.

Both of the packages you mentioned are listed as obsolete.

blankensteiner:
Indeed they are, my bad :-/

Any idea when you can create a release with this PR?

kerams:
I'd love to know that too; I am not a contributor/maintainer here :).

rwkarg:
Not sure where the HttpListener comment is from but `HttpListener` is a deprecated implementation and extremely inefficient compared to Kestrel.

kerams:
>extremely inefficient compared to Kestrel

Why would you need a high performance HTTP server for one endpoint that handles a single request every dozen seconds or so? Surely even despite its unofficial deprecation, HttpListener is a perfectly acceptable choice for a task like this.

blankensteiner:
Maybe also add netcoreapp3.1?

kerams:
Someone please correct me if I'm wrong, but I think since the list of dependencies for Core 3.0 and 3.1 targets will be identical, adding the latter has zero benefit.

JosephWoodward:
> Someone please correct me if I'm wrong, but I think since the list of dependencies for Core 3.0 and 3.1 targets will be identical, adding the latter has zero benefit.

Yes, that's my understanding.

PR #175, opened 2019-12-05
Fix metric middleware example

Marcel-B:
It should be LabelNames instead of Labels
A comma was missing


sandersaares:
Good catch! Thanks for the PR!

PR #178, opened 2019-12-16
Add endpoint routing extension method

### LINKED ISSUE #172: Endpoint routing
Going forward with ASP.NET Core 3.0, it looks like content-serving middleware should use [endpoint routing](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.0#endpoint-routing-differences-from-earlier-versions-of-routing).

Take [Health checks](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-3.0) for example: they're now registered via [MapHealthChecks](https://github.com/aspnet/AspNetCore/blob/658b37d2bd3b0da2c07e25b53fa5c97ce8d656d3/src/Middleware/HealthChecks/src/Builder/HealthCheckEndpointRouteBuilderExtensions.cs) instead of `UseHealthChecks` like was recommended in ASP.NET Core 2.2.

This would also allow for removal of the port parameter, as implementers could instead specify a hostname (or port) via the [RequireHost](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.routingendpointconventionbuilderextensions.requirehost?view=aspnetcore-3.0) extension method on the `IEndpointConventionBuilder` that the new `MapMetrics` extension method would return.

If there's interest, I'd be happy to contribute these changes.

chuseman:
Closes #172

### DIFF
--- a/Prometheus.AspNetCore/MetricServerMiddleware.cs
+++ b/Prometheus.AspNetCore/MetricServerMiddleware.cs
@@ -13,33 +13,18 @@ namespace Prometheus
     {
         public MetricServerMiddleware(RequestDelegate next, Settings settings)
         {
-            _next = next;
-
-            _onlyHandleRoot = settings.OnlyHandleRoot;
             _registry = settings.Registry ?? Metrics.DefaultRegistry;
         }
 
         public sealed class Settings
         {
-            public bool OnlyHandleRoot { get; set; } = true;
             public CollectorRegistry? Registry { get; set; }
         }
-
-        private readonly RequestDelegate _next;
-
-        private readonly bool _onlyHandleRoot;
 
         private readonly CollectorRegistry _registry;
 
         public async Task Invoke(HttpContext context)
         {
-            // We just handle the root URL (/metrics or whatnot).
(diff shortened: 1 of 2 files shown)


sandersaares:
Thank you for the PR!

It is not immediately obvious to me what is the purpose of the OnlyHandleRoot option - could you clarify why this is relevant here?

chuseman:
When the middleware registered via endpoint routing, `context.Request.Path.Value.Trim('/')` equals `"metrics"`, so the middleware wouldn't intercept the request. The new `OnlyHandleRoot` setting just disables that check so the request is handled by the middleware.

I thought it made sense to reuse the middleware, but maybe it would be simpler to have a separate class to use instead?

sandersaares:
I see what you mean. I think I will just get rid of the URL filtering altogether, as it is not really necessary and does not match usual middleware design patterns (where the router does the filtering, not the middleware itself).

sandersaares:
I see what you mean. I think I will get rid of the URL filtering altogether, as it is nonstandard middleware design and not actually necessary.

Issue #155, opened 2019-08-20
Ship additional net461 assembly?


chuseman:
We've got several older projects that still target .NET 4.6.1 and I'd like to bring in prometheus-net, but I'm unable to because of compilation issues.

For the unaware, there are known issues with .NET 4.6.1-4.7.1 for .NET Standard 2.0 references. [Microsoft's guidance](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting#multi-targeting) is to additionally target `net461` when targeting `netstandard2.0` as a workaround.

If there's interest, I'd be happy to contribute a PR for what I can. Specifically, I'd update the two `csproj` files as well as the `nuspec` files to copy both sets of libraries. Unfortunately a lot of the logic lives in the Azure DevOps Pipeline definitions for the Build and Release so I couldn't update those (unless you switch over to YAML configuration first).

sandersaares:
Do I understand it right that this would involve publishing different binaries, one for net461 and one for netstandard2.0? If so, I would not be willing to make this change.

The reason is that I deliberately try to keep the build and publishing pipeline minimally complex, as it is an easy place for "taxes" to incur that would have a negative effect on motivating quick publishing of fixes and updates (having to make and check different builds etc). As such, I do not accept changes into the library that would complicate the build process by outputting multiple sets of binaries (or by requiring multiple code paths for different targets).

While automation can help reduce such taxes, I would only consider this if targeting modern platforms that require such deviations (e.g. .NET Core 3.0, though I hope it will not) and not in case of legacy platforms that can (in theory) be upgraded from.

I would be interested to learn more about the specific issues you experienced - if not for else, then maybe to add to the readme so other users could be saved some effort in troubleshooting.

chuseman:
Trust me, I understand not wanting to bend over backward to support legacy platforms :) I wish it were an option for me.

Maybe it would be best if I describe exactly what I'm proposing. I'm just wanting an additional assembly in the `prometheus-net` NuGet package. It would live in the `lib/net461` folder in the package (next to the existing `lib/netstandard2.0` folder).

Multi-targeting is super easy now; you'd switch this line:

https://github.com/prometheus-net/prometheus-net/blob/d29ca370b1d1142eaea87f144e44ea8a486e3e9f/Prometheus.NetStandard/Prometheus.NetStandard.csproj#L4

To have multiple target frameworks (note the plural TargetFramework**s** node name):

```xml
[...]

PR #179, opened 2019-12-16
Add dependency for System.ValueTuple on .NET 4.6.1

chuseman:
In my previous .NET 4.6.1 PR #157, I added a PackageReference to System.ValueTuple as required for pre-4.7 builds but didn't add the appropriate dependency to the `.nuspec`, which lead to [this error](https://github.com/prometheus-net/prometheus-net/issues/155#issuecomment-560347625):

```
System.TypeInitializationException: The type initializer for 'Prometheus.Metrics' threw an exception. ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.ValueTuple, Version=4.0.3.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.
```

The workaround is to simply add the NuGet package reference yourself, but this dependency should fix it without having to manually add the reference.

### DIFF
--- a/Resources/Nuspec/prometheus-net.nuspec
+++ b/Resources/Nuspec/prometheus-net.nuspec
@@ -15,6 +15,9 @@
     <releaseNotes></releaseNotes>
 
     <dependencies>
+      <group targetFramework="net461">
+        <dependency id="System.ValueTuple" version="4.5.0" />
+      </group>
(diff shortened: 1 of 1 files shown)


sandersaares:
Thanks a lot!

PR #180, opened 2019-12-17
better use of concurrency in README example

r734:
It doesn't matter so much in this specific example, but `Thread.Sleep` is generally considered deprecated in favor of `Task.Delay` in concurrent applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,14 +82,14 @@ The following is a minimal implementation that simply increments a counter once 
 ```csharp
 using Prometheus;
 using System;
-using System.Threading;
+using System.Threading.Tasks;
 
 class Program
 {
     private static readonly Counter TickTock =
         Metrics.CreateCounter("sampleapp_ticks_total", "Just keeps on ticking");
 
-    static void Main()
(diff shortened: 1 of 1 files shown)


sandersaares:
I try to keep the examples deliberately "low tech" and since async main is still a relatively new feature in C#, it might confuse people not up to date with modern practices (despite async code being more relevant for modern apps, as you rightly imply). As such, I will not merge this at the moment.

There is definitely some room here to create more realistic examples for prometheus-net, of course. After all, not many apps just have a loop incrementing a counter. While the readme example is deliberately kept very oldschool, more realistic examples in the form of full runnable example project repos (or perhaps one repo containing multiple projects) would likely be a very welcome contribution if you feel like contributing more of your time.

r734:
Fair enough, though from my perspective, if I see `Thread.Sleep` in one of the first examples of a library, I question the quality of the rest of the library (I know that's not actually the case here--I'm just talking about first impressions).

Anyway, thanks for creating this library! It's very clean and well-documented aside from my little gripe :)

PR #189, opened 2020-01-16
Use the connection's cancellation token

blankensteiner:
When the request's connection is aborted, so should all operations.


sandersaares:

Issue #194, opened 2020-01-29
Actually target netstandard?


RheaAyase:
Hi all,

I don't know if I'm missing something but prometheus-net [nuget package](https://www.nuget.org/packages/prometheus-net/) seems to target only .NET Framework 4.61

More funny is that somehow `prometheus-net.AspNetCore` which does target .NET Standard seems to have `prometheus.net` as a dependency, while that one does not target .NET Standard.

If I'm missing something please explain how can one use these packages in .NET Core (on Linux)

RheaAyase:
Or perhaps the word target is not correct in this context, it simply does not list net standard.

sungam3r:
Interesting. Nuget package content seems normal:
![image](https://user-images.githubusercontent.com/21261007/73449015-44099480-4373-11ea-9247-55fec8350119.png)

sungam3r:
I think `prometheus-net.nuspec` file should have explicit dependencies groups for all targets.

sungam3r:
I checked it. If you do not use the nuspec file, but rely on `GeneratePackageOnBuild` property in csproj, then all dependency groups will be filled up. So I think we should modify nuspec. @sandersaares what do you think?

RheaAyase:
That's an interesting gotcha, I've never seen something like this happen.
Thanks for looking into it :D

sandersaares:
What exactly is the problem you are experiencing? Are you getting some error when attempting to reference the package?

sungam3r:
I think that the problem is that the [nuget page](https://www.nuget.org/packages/prometheus-net/) misleads the reader (it lists only net461 target):
![изображение](https://user-images.githubusercontent.com/21261007/74101799-b5ed9500-4b4e-11ea-8c52-ccac464cf71e.png)

sandersaares:
I see. Okay - while I don't think relying on the dependencies list to determine supported frameworks is really the intended usage of that list, I suppose it won't hurt to be more explicit and list them all, though, to avoid confusion on this topic.

sandersaares:
![image](https://user-images.githubusercontent.com/9914262/74102828-d66a1f00-4b4f-11ea-981d-e318e6d6ae9e.png)

There we go, I hope that helps!

PR #195, opened 2020-01-30
Add netcoreapp3.1 target

sungam3r:


johnnyreilly:
Thanks so much for raising this PR!  Just spent the morning wondering why things weren't working and I suspect this is the reason 😄

sungam3r:
@sandersaares Any progress on this?

sungam3r:
Relates to #194 . @sandersaares Do I need to also add dependencies groups in nuspec?

sandersaares:
What exactly does the change in this PR accomplish? The library is already compatible with ASP.NET Core 3.1 as far as I know. Are you having some problem using it with ASP.NET Core 3.1, @sungam3r?

sandersaares:
@johnnyreilly the stack trace you pasted indicates that you are not using prometheus-net but rather the Prometheus integration in App Metrics, which is a different library altogether and not related to prometheus-net.

sungam3r:
> What exactly does the change in this PR accomplish? The library is already compatible with ASP.NET Core 3.1 as far as I know. Are you having some problem using it with ASP.NET Core 3.1, @sungam3r?

No problems yet. I migrate the backend to 3.1 and noticed that this package targets 3.0. I thought that it would not hurt to explicitly specify the 3.1 target as well especially since it is LTS.

kerams:
>I thought that it would not hurt to explicitly specify the 3.1 target as well

Well, it's going to increase the Nuget package size while providing no benefit at all.

sandersaares:
Since the present state of things does not lead to any issues, I will not merge this PR that would include both 3.0 and 3.1 binaries in the package, as it would not produce any benefit.

It is true that .NET Core 3.0 is end-of-life in approximately 1 month. However, on a technical level that does not strictly matter for prometheus-net given that you can install it fine into a 3.1 project even if it targets 3.0 internally. I imagine in a future package version the target will change to 3.1 instead of the current 3.0, to track the supported .NET Core versions.

johnnyreilly:
@sandersaares you were right; this is actually the issue I'm facing: https://github.com/AppMetrics/AppMetrics/issues/396

PR #202, opened 2020-04-01
Set thread names

macaba:
(.NET 5 Preview 1)
Using `htop` on Linux, and switching to tree view & enabling custom thread names will show you all the child thread names:

image

This pull request will show `Metric Listener` and `Metric Process` as child thread names.

### DIFF
--- a/Prometheus.NetStandard/MetricServer.cs
+++ b/Prometheus.NetStandard/MetricServer.cs
@@ -35,6 +35,7 @@ namespace Prometheus
             {
                 try
                 {
+                    Thread.CurrentThread.Name = "Metric Server";     //Max length 16 chars (Linux limitation)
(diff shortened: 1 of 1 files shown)


sandersaares:

Hmmm, in principle nothing guarantees that these tasks run in dedicated threads. But of course in practice all known implementations do equate `TaskCreationOptions.LongRunning` to creating a dedicated thread so perhaps that's okay. I'm sure enough code in the world relies on this for it to be a de-facto requirement by now, anyway.

macaba:
True! I agree that LongRunning is now de facto syntax sugar as a short path to ‘new Thread’ under the hood. Thanks for merging, I look forward to seeing the thread names in production.

sandersaares:

PR #208, opened 2020-04-24
Reducing metricpussher allocations

kripergvg:
I found out that MetricPusher wasted memory because he created new MemoryStream for every push.
In this PR I use RecyclablememoryStream https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream
 Benchmark result
Before 
image
After 
image
The benchamrk in MetricPusherBenchmarks.cs

### DIFF
--- a/Benchmark.NetCore/MetricPusherBenchmarks.cs
+++ b/Benchmark.NetCore/MetricPusherBenchmarks.cs
@@ -1,0 +1,58 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Engines;
+using Prometheus;
+
+namespace Benchmark.NetCore
+{
+    [MemoryDiagnoser]
+    [SimpleJob(RunStrategy.Monitoring, warmupCount:0)]
+    public class MetricPusherBenchmarks
+    {
+        [Benchmark]
+        public async Task PushTest()
+        {
+            var registry = Metrics.NewCustomRegistry();
+            var factory = Metrics.WithCustomRegistry(registry);
+            
+            var pusher = new MetricPusher(new MetricPusherOptions
+            {
+                Endpoint = "http://127.0.0.1:9091/metrics",
+                Registry = registry,
+                IntervalMilliseconds = 30,
+                Job = "job"
+            });
+            pusher.Start();
+
+            var counters = new List<Counter>();
+            for (int i = 0; i < 1000; i++)
+            {
+                var counter = factory.CreateCounter($"Counter{i}", String.Empty);
+                counters.Add(counter);
+            }
+
+            var ct = new CancellationTokenSource();
+            var incTask = Task.Run(async () =>
+            {
+                while (!ct.IsCancellationRequested)
+                {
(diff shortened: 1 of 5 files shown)


sandersaares:
I am having some trouble reproducing your benchmarks using the code you provided. It looks like the benchmark code assumes that there is an actual metricpusher listening? Otherwise the GCs do not occur because the connection just fails and the runtime has an easier time.

With a dummy endpoint (random HTTP server), I do not see any differences in the benchmarks before/after: 

```
|   Method |    Mean |    Error |   StdDev |     Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------- |--------:|---------:|---------:|----------:|----------:|------:|----------:|
| PushTest | 5.073 s | 0.0225 s | 0.0149 s | 2000.0000 | 1000.0000 |     - |   9.54 MB |

|   Method |    Mean |    Error |   StdDev |     Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------- |--------:|---------:|---------:|----------:|----------:|------:|----------:|
| PushTest | 5.193 s | 0.0167 s | 0.0111 s | 2000.0000 | 1000.0000 |     - |   9.54 MB |
```

Please see if you can adjust the benchmarks to more accurately reflect the experience you have in your runtime environment. If they need a web API (e.g. pusher) to be listening, that web API should be part of the benchmark infrastructure.

sandersaares:
If I run a pushgateway, I do see the benefit. Though this makes me wonder - do we even need that MemoryBuffer there at all? I will investigate if this can be done without any buffering. The MemoryBuffer is currently there to avoid taking some dependencies but maybe it can be eliminated.

sandersaares:
Okay looks like I can achieve the same results with a no-buffering implementation:

```
Baseline
|   Method |    Mean |    Error |   StdDev |     Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------- |--------:|---------:|---------:|----------:|----------:|------:|----------:|
| PushTest | 5.049 s | 0.0220 s | 0.0145 s | 6000.0000 | 1000.0000 |     - |  28.28 MB |

Recycling MemoryBuffer
|   Method |    Mean |    Error |   StdDev |     Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------- |--------:|---------:|---------:|----------:|----------:|------:|----------:|
| PushTest | 5.056 s | 0.0342 s | 0.0226 s | 3000.0000 | 1000.0000 |     - |  15.91 MB |

No buffering
|   Method |    Mean |    Error |   StdDev |     Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------- |--------:|---------:|---------:|----------:|----------:|------:|----------:|
| PushTest | 5.059 s | 0.0264 s | 0.0175 s | 4000.0000 | 1000.0000 |     - |  15.98 MB |
```

So that lends some credibility to the benchmarks being "good enough". I will commit the no-buffering implementation as it avoids taking a dependency on the recycling buffer library. I will adjust the benchmark so that it expects the Tester app to be running and providing a dummy pushgateway - not as clean as I would like but good enough and it avoids the "receiving" side polluting benchmark data.

sandersaares:
I manually merged the benchmark code and integrated a no-buffer implementation instead. Thanks for the inspiration!

PR #212, opened 2020-05-11
Implement gRPC metrics

### LINKED ISSUE #211: gRPC metrics
I noticed that this project has some standard HTTP metrics (via `UseHttpMetrics`), but would it be possible to add gRPC metrics too? For example, I'd like to see the number of calls to each of my gRPC services. Probably something like the `http_requests_received_total` counter but with service and method name instead of controller and action.

Daniel15:
This is an initial implementation of gRPC metrics, based off the existing HTTP metrics code.

Currently this just adds one metric, a `grpc_requests_received_total` metric:
```
# HELP grpc_requests_received_total Provides the count of gRPC requests that have been processed
# TYPE grpc_requests_received_total counter
grpc_requests_received_total{service="Greet.Greeter",method="SayHello"} 2
```

Since this code has a dependency on the gRPC assembly (`Grpc.AspNetCore.Server`), I created a separate `Prometheus.Grpc` assembly for the middleware.

Closes #211

### DIFF
--- a/Benchmark.NetCore/Benchmark.NetCore.csproj
+++ b/Benchmark.NetCore/Benchmark.NetCore.csproj
@@ -2,13 +2,11 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
-    <LangVersion>latest</LangVersion>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
     <SignAssembly>true</SignAssembly>
 
     <AssemblyOriginatorKeyFile>..\Resources\prometheus-net.snk</AssemblyOriginatorKeyFile>
 
-    <LangVersion>latest</LangVersion>
     <WarningsAsErrors />
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
@@ -22,7 +20,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="BenchmarkDotNet" Version="0.11.5" />
+    <Compile Include="..\Tester.NetFramework\TesterConstants.cs" Link="TesterConstants.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="0.12.1" />
   </ItemGroup>
 
   <ItemGroup>
--- a/Benchmark.NetCore/MetricPusherBenchmarks.cs
+++ b/Benchmark.NetCore/MetricPusherBenchmarks.cs
@@ -1,0 +1,86 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Engines;
+using Prometheus;
+using tester;
+
+namespace Benchmark.NetCore
+{
+    // NB! This benchmark requires the Tester project to be running and the MetricPusherTester module to be active (to receive the data).
+    // If there is no tester listening, the results will be overly good because the runtime is under less I/O load.
+    [MemoryDiagnoser]
+    [SimpleJob(RunStrategy.Monitoring, warmupCount:0)]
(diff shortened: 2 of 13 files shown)


Prometheus.Grpc/GrpcMiddlewareExporterOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
```
sungam3r, line 3: Please cleanup not used usings.

Prometheus.Grpc/Prometheus.Grpc.csproj
```diff
@@ -0,0 +1,53 @@
...
+    <DocumentationFile>bin\Release\netstandard2.0\Prometheus.Grpc.xml</DocumentationFile>
+    <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
+    <NoWarn>1701;1702;1705;1591</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <DocumentationFile>bin\Debug\netstandard2.0\Prometheus.Grpc.xml</DocumentationFile>
+    <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
+    <NoWarn>1701;1702;1705;1591</NoWarn>
+  </PropertyGroup>
```
sungam3r, line 29: This can be merged in single PropertyGroup

Daniel15: I intentionally kept the same style as `Prometheus.AspNetCore.csproj` because I wasn't sure of the coding standards in this project. I can change it though.

Daniel15:
I can't work out how to get Azure Pipelines to use the .NET Core 3 build of `Tests.NetCore.dll` rather than .NET Core 2. I can't seem to find the config for it.

albertyfwu:
@Daniel15 Are you still planning to continue this work?

Daniel15:
Sure. I just need to clean up the csproj file based on @sungam3r's comment, and figure out how to change the Azure Pipelines build to use the .NET Core 3 build of the tests. If someone could help me with the latter, I'd appreciate it.

sandersaares:
The Azure Pipelines runner deliberately targets .NET Core 2 when running tests because the ASP.NET Core version target is 2.1.

If you intend the gRPC part to only support .NET Core 3 then the tests for it should be in a separate test project (e.g. Tests.NetCore3). The version target should be 3.1 then (let's target LTS; I see old tests project has 2.0 in it, probably just historical and nobody updated).

In fact, I will just go ahead and do this test runtime split in master & pipeline to help things along.

Prometheus.Grpc/Prometheus.Grpc.csproj
sandersaares: The project file ugly PropertyGroups are often ugly because they are made like that by Visual Studio in some cases. Feel free to make it pretty - there is no strict convention to keep it ugly.

sandersaares:
After a second pass, I will just upgrade the tests project to target 3.1 in master and take the risk that this will leave uncovered some 2.x issues, in the name of easier development.

Daniel15:
> If you intend the gRPC part to only support .NET Core 3

gRPC itself only supports .NET Core 3 (see https://www.nuget.org/packages/Grpc.AspNetCore) so any code that uses it also requires .NET Core 3, which is why I used it here instead of sticking with the same .NET version as everything else.

sandersaares:
Makes sense.

The code looks pretty solid and I am happy to merge it.

To make development & maintenance easy, I would like to also see a "Tester" app implementation included in the solution. Something which would allow a developer to step through the logic in the context of a real gRPC request being handled. Equivalent to `AspNetCoreMiddlewareTester`, basically.

Do you have the availability to add this?

sandersaares:
OK, I'll just add the tester myself and merge this.

sandersaares:
Thanks for the contribution! Feel free to extend it further when you have the time for it.

sandersaares:

Daniel15:
Thank you for merging this! Sorry I didn't have time to add the tester myself.

PR #216, opened 2020-05-14
close response stream in AspNetMetricServer

### LINKED ISSUE #215: AspNet NetFramework implementation hangs
So I have configured the route in my `Global.asax` using `   AspNetMetricServer.RegisterRoutes(GlobalConfiguration.Configuration)` as defined in the docs. I have added a counter to an endpoint, incrementing the counter on each successful processing at the endpoint, as per the docs. However, all requests to the `/metrics` on this host just hang and never gets a response. This happens both when configured inside prometheus as a scrape target and when accessing the `/metrics` url directly. (I used fiddler to inspect the web requests and I can see they are not getting any response).

If I try access the `/metrics` url, all subsequent requests to other endpoints also hang, and after x requests by prometheus scraper, all subsequent requests to other endpoints also hang.

I took the source code for AspNetMetricServer and modified the content of the response to simply be a StringContent hello and it returns fine, so it seems the issue is within the `CollectAndExportAsTextAsync` method, unless there is some weird config I am missing. 

[update]
After more investigation, it appears that this is an issue with the stream not being flushed. The headers are returned on the first request, but the body is never returned. 

[update 2]
After further investigation it seems the underlying implementation of the `PushStreamContent` class is a `CompleteTaskOnCloseStream` class which does not flush, hence closing of the stream is required. This does not actually close the underlying stream though because you are not the owner, see the implementation [here](https://github.com/aspnet/AspNetWebStack/blob/master/src/System.Net.Http.Formatting/PushStreamContent.cs). Hence in method `FlushAsync` in file `TextSerializer.cs`, after calling `FlushAsync` on the stream, you need to call `_stream.Value.Close();` which completes the task, hence returning the body of the request.

steven-dawkins:
resolves #215 

this could probably do with a unit test but I am discouraged by this line: https://github.com/prometheus-net/prometheus-net/blob/ee2bdbd3330548059fde9eb12677c35401b527b6/Tester.NetFramework/KestrelMetricServerTester.cs#L9

### DIFF
--- a/Prometheus.NetFramework.AspNet/AspNetMetricServer.cs
+++ b/Prometheus.NetFramework.AspNet/AspNetMetricServer.cs
@@ -47,9 +47,16 @@ namespace Prometheus
 
                 var response = new HttpResponseMessage(HttpStatusCode.OK)
                 {
-                    Content = new PushStreamContent((stream, content, context) =>
+                    Content = new PushStreamContent(async (stream, content, context) =>
                     {
-                        return Metrics.DefaultRegistry.CollectAndExportAsTextAsync(stream, default);
+                        try
+                        {
+                            await Metrics.DefaultRegistry.CollectAndExportAsTextAsync(stream, default);
+                        }
+                        finally
+                        {
+                            stream.Close();
+                        }
                     }, PrometheusConstants.ExporterContentTypeMinimal)
                 };
 


lukethompsxn:
This looks good and is a more ideal fix than the downstream fix I currently have which would impact other versions too.

sandersaares:
Good catch, good fix. Thanks a lot!

sandersaares:

Issue #225, opened 2020-06-22
Include RoutePattern and Path in http_request_* metrics


tpodolak:
Current implementation of HTTP like metrics for ASP.NET Core dump labels for code,method,controller and action.
```text
http_request_duration_seconds_sum{code="401",method="GET",controller="Weather",action="Get"} 0.1214422
http_request_duration_seconds_count{code="401",method="GET",controller="Weather",action="Get"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.001"} 8
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.002"} 16
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.004"} 18
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.008"} 20
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.016"} 20
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.032"} 20
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.064"} 20
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.128"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.256"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="0.512"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="1.024"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="2.048"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="4.096"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="8.192"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="16.384"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="32.768"} 21
http_request_duration_seconds_bucket{code="401",method="GET",controller="Weather",action="Get",le="+Inf"} 21
```

Would it be possible to also add request path and route template to the labels?
[...]

PR #228, opened 2020-07-11
GH-225 - including route pattern in labels

tpodolak:
Includes route pattern in HttpMetrics. Please note that pattern can be only intercepted when the API is configured to UseEndpointRouting which is not available prior to ASP.NET Core 2.2. Additionally, I've changed casing of `Tester.NetFramework/tester.netframework.csproj` file, so now it is possible to open it on Linux-like systems

Closes https://github.com/prometheus-net/prometheus-net/issues/225


Prometheus.AspNetCore/Prometheus.AspNetCore.csproj
```diff
@@ -37,7 +37,7 @@
   </ItemGroup>
   
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.2.0" />
```
tpodolak, line 40: IEndpointFeature and in general endpoint routing is only available from ASP.NET Core 2.2. As I wasn't sure if you want to keep the ASP.NET version in sync across all projects, I've updated it everywhere

sandersaares: The code intentionally targets 2.1 because that is the v2 LTS version, with 2.2 being end of life (ASP.NET Core and .NET Core are versioned together).

If the minimum requirement is 2.2 for this then this feature should target 3.1 instead (with an `#if NETCOREAPP3_1`).

sandersaares:
Beyond the ASP.NET Core version topic, this looks good to me. However, see https://github.com/prometheus-net/prometheus-net/issues/225#issuecomment-657384161 - if this is not immediately required, I will defer merging it until there is repeated interest, as it feels like a feature not many users would be interested in (but also one that might generate needless labels for all users and require maintenance later down the line).

Prometheus.AspNetCore/Prometheus.AspNetCore.csproj
tpodolak: Reverted 2.2 version bump in favor of `#if NETCOREAPP3_1`

tpodolak:
>if this is not immediately required, I will defer merging it until there is repeated interest,
I am fine with deferring the merge. I am sure that sooner or later it will get more votes from the community

JosephWoodward:
We need something similar to this (not every route maps to a controller/action, take Razor pages for example), however I can see an issue with this. As it creates a label value from the route data, it runs the risk of creating a lot of metrics if an address contains unique data such as a guid.

altinoren:
> 
> 
> ... if this is not immediately required, I will defer merging it until there is repeated interest, as it feels like a feature not many users would be interested in (but also one that might generate needless labels for all users and require maintenance later down the line).

I'd love to have this in future releases, to handle versioned endpoints. It's common in API scenarios to have multiple actions with the same name, only to be differentiated by an `api-version` parameter or something similar.

sandersaares:
Support for adding custom parameters is already merged into 4.0.0. The deferred feature is exposing the raw template string itself. If all you need is to have the parameter present in labels, 4.0.0 will serve you nicely.

Version 4.0.0 will be published later today, unless some last minute validation issues come up.

csseba:
Hi @sandersaares @tpodolak ,

 I am using this lib  version and in the startup.cs I have following code:

`app.UseHttpMetrics(options =>
            {
                options.AddRouteParameter("route");
            });`
            
However, when I get metrics, I'm getting route = ""
`http_request_duration_seconds_bucket{code="404",method="GET",controller="TestController",action="Get",route="",le="0.016"} 0`

What should I do in order to have the route correctly mapped?

I am using .net core 3.1 and my controllers classes are like:
`[Route("v2.0/products-test")]`

I think I misunderstand something.

Thanks in advance

sandersaares:
There is now the possibility to inspect the HTTP request to add custom labels, so I think all the requested use cases should be covered now. Will close this.

PR #231, opened 2020-07-27
fixed link text

KinNeko-De:
Hello,
I only fixed the name of the link because it confused while installing the libraries.

Greetings


sandersaares:
Thank you for the contribution! Good catch!

PR #232, opened 2020-07-28
Added host label to http request metrics

LeoHexspoor:
It's common for multi-tenant applications to use the domain as a tenant identifier. By adding the host as a label multi-tenant applications using this mechanism can now also view HTTP metrics on a per tenant basis.

### DIFF
--- a/Prometheus.AspNetCore/HttpMetrics/HttpRequestLabelNames.cs
+++ b/Prometheus.AspNetCore/HttpMetrics/HttpRequestLabelNames.cs
@@ -9,19 +9,22 @@ namespace Prometheus.HttpMetrics
         public const string Method = "method";
         public const string Controller = "controller";
         public const string Action = "action";
+        public const string Host = "host";
 
         public static readonly string[] All =
         {
             Code,
             Method,
             Controller,
-            Action
+            Action,
+            Host
         };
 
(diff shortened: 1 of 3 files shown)


sandersaares:
Makes sense. I wonder if there are also additional properties of the HTTP request that might be beneficial to expose? Perhaps some general purpose form like `options.AddRequestLabel("has_my_special_token", request => request.Headers.ContainsKey("Contoso-Token"))`

sandersaares:
For the moment I will defer merging this, in favor of the more general approach. Any comments you have on that design are appreciated!

LeoHexspoor:
This could indeed be made more generic, I'm assuming the intention of your design would be that the function returns a string and not a boolean?  And that this return value is used as the label value?

I can image you may also want to provide an option with two functions; one to check if you want to output the label using a boolean return value and the second function to return the value that is used as the label value.

sandersaares:
> I'm assuming the intention of your design would be that the function returns a string and not a boolean? And that this return value is used as the label value?

Right, sorry for the confusing boolean example - just a mistake there.

sandersaares:
There is now `options.AddCustomLabel()` available in v6.0.0 preview version which enables this scenario, as well as other scenarios not currently foreseen. Comments most welcome!

I will close this PR as the new generic feature supersedes this special-case feature.

PR #236, opened 2020-08-18
Ignore http metrics by condition

Sverdel:
In my case I have both gRpc and REST endpoints in my app. So, I don't want to write http metrics when processing gRPC request. With this changes I can filter gRPC requests, for example, by request contentType header. Something like this:
```
app.UseHttpMetrics(o =>
{
    o.SetIgnoreCondition(ctx => ctx.Request.ContentType == "application/grpc");
});
```


Sverdel:
@sandersaares, if you have any doubts about this PR, please, let me know

sandersaares:
I agree it would be nice if the two metric types (gRPC/HTTP) were mutually exclusive out of the box, I think filtering functionality belongs to ASP.NET Core, not prometheus-net.

Wouldn't using a conditional mapping satisfy the scenario? Something like this:

```csharp
app.UseWhen(context => context.Request.ContentType != "application/grpc", config =>
{
    config.UseHttpMetrics();
});`
```

PR #242, opened 2020-09-02
Update the doc to mention aspnetcore 3.1

gldraphael:
While it could, it doesn't work with 3.0 if the target-framework is set as `netcoreapp3.0`. 
See: https://github.com/prometheus-net/prometheus-net/commit/66ec9ebdc7f1a792fdcd6fa910aff6b6ee5c44b8


sandersaares:

PR #248, opened 2020-09-14
Httpclient

AChehre:


Prometheus.HttpClient/HttpClientMetrics/HttpClientRequestLabelNames.cs
```diff
@@ -0,0 +1,31 @@
...
+        public static IDictionary<string, string> AsParameterMap()
+        {
+            var map = new Dictionary<string, string>(All.Length);
+
+            foreach (var item in All)
+            {
+                map[item] = item;
+            }
+
+            return map;
```
NazmiAltun, line 28: This function can be simplified with  LINQ ToDictionary()

sandersaares:
I am thinking how to allow users to tell the difference between HTTP client instances. The HttpClientProvider pattern has both named clients and typed clients which provide useful data. Unfortunately, this information does not appear to be exposed to the middleware that we can inject into the pipeline.

```csharp
services.AddHttpClient("foo", client =>
{
    client.DefaultRequestHeaders.Add("foo", "bar");

}).UseHttpClientMetrics("foo");
```

We could do something like this, although I am not a fan of having to repeat the "foo" name in two places. Any thoughts, @AChehre ? Do you see a way around this repetition to allow the specific client to be identified?

sandersaares:
I have merged this pull request manually after making some minor adjustments to simplify the structure (no separate projects for HttpClient) and unifying code style with existing.

The question of how to tell the difference between different HttpClients still remains but we can leave it for version 2.

Thank you very much for this great contribution! I will mark this PR as closed, since it is already merged manually.

Issue #246, opened 2020-09-14
Why /metrics endpoint return text not in prometheus format?


wujun4code:
## what i have done?

```c#
app.UseHttpMetrics();
app.UseMetricServer();
```
## what happend?

When I began to debug it locally, the localhost:5000/metrics return the following content:

```json
{
  "contexts": [
    {
      "apdexScores": [
        {
          "frustrating": 0,
          "sampleSize": 0,
          "satisfied": 0,
          "score": 1,
          "tolerating": 0,
          "name": "Apdex|server:myserver,app:myweb,env:development",
          "tags": {
            "server": "myserver",
            "app": "mywe",
            "env": "development"
          }
        }
      ],
      "context": "Application.HttpRequests",
      "counters": [],
      "gauges": [],
      "histograms": [],
      "meters": [],
      "timers": [
        {
          "activeSessions": 0,
          "count": 0,
          "durationUnit": "ms",
          "histogram": {
            "lastValue": 0,
            "max": 0,
            "mean": 0,
            "median": 0,
            "min": 0,
            "percentile75": 0,
            "percentile95": 0,
            "percentile98": 0,
            "percentile99": 0,
            "percentile999": 0,
            "sampleSize": 0,
            "stdDev": 0,
            "sum": 0
          },
          "rate": {
            "fifteenMinuteRate": 0,
            "fiveMinuteRate": 0,
            "meanRate": 0,
            "oneMinuteRate": 0
          },
          "rateUnit": "min",
          "name": "Transactions|server:myserver,app:imywe,env:development",
          "tags": {
            "server": "myserver",
            "app": mywe",
            "env": "development"
          },
          "unit": "req"
        }
      ]
    },
    {
      "apdexScores": [],
      "context": "Application",
      "counters": [],
      "gauges": [],
      "histograms": [],
      "meters": [],
      "timers": []
    }
  ],
  "timestamp": "2020-09-14T08:40:28.5164049Z"
}
```

Issue #214, opened 2020-05-12
Prometheus-net can't reach metrics on kubernetes


ER-git:
Hey I wrote .net core application (core 3 not ASP.net).
And im using the following:
```
            MetricServer server = new MetricServer(hostname: "localhost", port: MetricsPort);
            server.Start();
```
Also tried 127.0.0.1 instead of localhost.
on windows everything is running great , but when I try to reach the address on kubernetes(using node-port) I receive XXX.XXX.X.XX refused to connect.
Docker file got expose param as follow:
```
FROM mcr.microsoft.com/dotnet/core/aspnet:3.1-bionic AS base
WORKDIR /app

EXPOSE 9090
EXPOSE 30000
EXPOSE 8080

FROM mcr.microsoft.com/dotnet/core/sdk:3.1-bionic AS build
WORKDIR /src

COPY XXXXX
COPY XXXX
COPY XXXXX
RUN dotnet restore XXXX
COPY . .
WORKDIR /src/XXX
RUN dotnet build XXXX -c Release -o /app

FROM build AS publish
RUN dotnet publish XXXX -c Release -o /app

FROM base AS final
WORKDIR /app
COPY --from=publish /app .
EXPOSE 9090
EXPOSE 30000
EXPOSE 8080
ENTRYPOINT ["dotnet", "XXX.dll"]
```
And the target port where the MetricsPort is on 9090

sandersaares:
This sounds like a question of container network communications design and configuration and therefore out of scope for an issue in the repository of the prometheus-net library. As such, I close this issue.

ChristopherLenz:
@ER-git were you able to solve it. I have a similar problem, which is really strange. I have roughly 10 .net Core APIs running. Everyone can be reached in Kubernetes. Except this one which only purpose is to expose some metrics.

PR #251, opened 2020-09-29
Add note on exposing metrics server outside localhost

Starefossen:
We noticed that we were unable to reach the metrics server from our Prometheus server in out Kubernetes environment. We had copied the example from the README where the metrics server which binds the `hostname` to `localhost`. Changing the it to `*` solved our problem.

This pull request adds a note to the README.md regarding this.

Related #246 #214

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,8 @@ class Program
 }
 ```
 
+**NB!** Keep in mind that if you want to expose the metrics server outside of the local network you need to set the `hostname` to `*`.
+
 # Counters
 
 Counters only increase in value and reset to zero when the process restarts.


sandersaares:
I understand the problem you face. The reason that the quick start uses `localhost` is to enable a maximally-successful example, one that does not need any special configuration. Listening on `*`, unfortunately does require some permissions configuration (or run as Administrator).

You are right to say that this is a bit misleading, however. I will merge a slightly modified version of what you propose here, referencing the chapter with the permissions configuration instructions.

sandersaares:
I have merged the modified variant of this PR. Thank you for the contribution!

Starefossen:
Thank you @sandersaares. I believe this will benefit many users and hopefully save them some time debugging 😄 Thanks for the awesome work you are doing with this library ❤️

PR #264, opened 2020-10-31
Conditional TargetFrameworks in Prometheus.NetStandard

JessThrysoee:
To be able to build on linux and macos


sandersaares:
I can see this making sense to include. Thank you for your contribution!

PR #266, opened 2020-11-01
Await SendAsync to record correct duration

### LINKED ISSUE #265: HttpClientRequestDurationHandler not recording correct duration
### Steps to reproduce

1. Create an empty dotnet web project
2. Use `HttpClientMetrics()` and `MapMetrics()`
```c#
    services.AddHttpClient<FooApi>()
        .UseHttpClientMetrics();
    ...
    
    app.UseEndpoints(e => e.MapMetrics())
```
3. Exercise the HttpClient, and compare metrics from the console logs vs /metrics

### Expected

* Console log timing should be similar to /metrics

### Actual

* Metrics is much lower than expected

---

`HttpClientRequestDurationHandler` is returning the Task directly without awaiting for the request task to complete. This means the timer is disposed as soon as the method is returned and is actually recording method invocation time, not actual client request duration.

kenkam:
Fixes #265 

Await `base.SendAsync()` in `HttpClientRequestDurationHandler` to correctly record the request duration

### DIFF
--- a/Prometheus.AspNetCore/HttpClientMetrics/HttpClientRequestDurationHandler.cs
+++ b/Prometheus.AspNetCore/HttpClientMetrics/HttpClientRequestDurationHandler.cs
@@ -11,11 +11,11 @@ namespace Prometheus.HttpClientMetrics
         {
         }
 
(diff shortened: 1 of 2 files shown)


sandersaares:
Good catch! Thank you for the contribution!

PR #270, opened 2020-11-24
Added option to add custom labels for http metrics

mvSapphire:
Added ability to add custom labels to HTTP metrics

### DIFF
--- a/Prometheus.AspNetCore/HttpMetrics/HttpMetricsOptionsBase.cs
+++ b/Prometheus.AspNetCore/HttpMetrics/HttpMetricsOptionsBase.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
 
 namespace Prometheus.HttpMetrics
 {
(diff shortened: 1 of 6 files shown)


sandersaares:
A similar feature in a more general form is now available in prerelease build `4.1.0-pre-000171-15be8f3` in the form of static labels: https://github.com/prometheus-net/prometheus-net#static-labels

Please give it a try and let me know if it satisfies your needs in this scenario.

mvSapphire:
> A similar feature in a more general form is now available in prerelease build `4.1.0-pre-000171-15be8f3` in the form of static labels: https://github.com/prometheus-net/prometheus-net#static-labels
> 
> Please give it a try and let me know if it satisfies your needs in this scenario.

the idea of this feature to add values for these labels on runtime. 

but in the documentation, static labels are labels that always have fixed values. that not what we actually need

sandersaares:
I see what you mean - you want the custom label values to differ based on the data in the request? Indeed, that is a different use case. I will keep this in the backlog for a bit - it requires some thinking to see how to most safely do this, as it feels like something that could have a lot of corner cases.

pinowthebird:
@sandersaares Any update on this? It might make sense to list the corner cases.

I have a similar use case which could definitely be solved by this PR. Adding custom labels on the current HTTP metrics would be an awesome addition.

lipw2019:
I was confused by the readme AddCustomLabel, If it's not in the release, would be better to remove it.
https://github.com/prometheus-net/prometheus-net/blob/c81d12d43cc53081155578ed8a915fce88c74f0f/README.md?plain=1#L445

sandersaares:
AddCustomLabel is currently available in the latest prerelease package. It will be available shortly in a stable version, as well.

PR #276, opened 2020-12-08
Save allocation of Stopwatch in HttpRequestDurationMiddleware

agehrke:
The `ValueStopwatch` struct mimics the behavior of `Stopwatch` class. The implementation is copied directly from [dotnet/extensions source code](https://github.com/dotnet/extensions/blob/master/src/Shared/src/ValueStopwatch/ValueStopwatch.cs).


agehrke:
I see that there's also potential in `MetricPusher.StartServer` and its while(true) loop.

sandersaares:
Very sensible - thank you for the contribution!

PR #296, opened 2021-04-13
csproj updates for net5.0 support

### LINKED ISSUE #267: Unintended PackageReference to Microsoft.AspNetCore when targeting net5.0
The project file logic [here](https://github.com/prometheus-net/prometheus-net/blob/master/Prometheus.AspNetCore/Prometheus.AspNetCore.csproj#L35-L37) results in a PackageReference to the old aspnetcore nuget packages when using this library in a net5.0 application.

I believe it should be as simple as changing the logic here to be the inverse of the condition [here](https://github.com/prometheus-net/prometheus-net/blob/master/Prometheus.AspNetCore/Prometheus.AspNetCore.csproj#L39), which would allow for some flexibility on new runtimes.

Alternatively, the library could be updated to multitarget to net5.0 and do a similar condition for net5.0.

In any case, I'd be happy to contribute this change.

baronfel:
Fixes #267 by adding an additional multitarget for net5.0 whereever we had a prior one for netcoreapp3.1.

This results in packages with TFM groups for

* netstandard2.0
* netcoreapp3.1
* net5.0

The addition of the explicit net5.0 group prevents .net5.0-targeting consumers from falling-back to the netstandard2.0 dependencies, because net5.0 does not fallback to netcoreapp3.1 in nuget-TFM-parlance.

In general, I had to do the same kinds of operations through this work:

* add the new net5.0 TFM
* check additional dependencies in the project to key the use of specific versions off of which TFM was being targeted
* scope the variations to be a tightly-scoped as possible, to make the full set of changes easy to reason about.

After this work, the nuget packages expose the following dependency graph:

image

Which looks perfect to me. No net5.0 TFMs have 3.x dependencies or AspNetCore 2.x dependencies, and no netcoreapp3.1/netstandard2.0 TFMs have any net50 dependencies.

### DIFF
--- a/Prometheus.AspNetCore.Grpc/Prometheus.AspNetCore.Grpc.csproj
+++ b/Prometheus.AspNetCore.Grpc/Prometheus.AspNetCore.Grpc.csproj
@@ -1,10 +1,11 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
     <RootNamespace>Prometheus</RootNamespace>
 
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
 
     <SignAssembly>true</SignAssembly>
 
@@ -16,22 +17,13 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
     <NoWarn>1701;1702;1705;1591</NoWarn>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <DocumentationFile>bin\Release\netstandard2.0\Prometheus.Grpc.xml</DocumentationFile>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <DocumentationFile>bin\Debug\netstandard2.0\Prometheus.Grpc.xml</DocumentationFile>
+    <IsPackable>true</IsPackable>
+    <OutputType>Library</OutputType>
   </PropertyGroup>
 
   <ItemGroup>
     <Compile Include="..\Resources\SolutionAssemblyInfo.cs" Link="SolutionAssemblyInfo.cs" />
-    <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Grpc.AspNetCore.Server">
-      <Version>2.28.0</Version>
-    </PackageReference>
+    <PackageReference Include="Grpc.AspNetCore.Server" Version="2.32.0"/> <!-- first version that has dedicated net5.0 support -->
     <ProjectReference Include="..\Prometheus.AspNetCore\Prometheus.AspNetCore.csproj" />
   </ItemGroup>
 
--- a/Prometheus.AspNetCore.HealthChecks/Prometheus.AspNetCore.HealthChecks.csproj
+++ b/Prometheus.AspNetCore.HealthChecks/Prometheus.AspNetCore.HealthChecks.csproj
@@ -1,10 +1,11 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
 	<PropertyGroup>
-		<TargetFrameworks>netstandard2.0;netcoreapp3.1</TargetFrameworks>
+		<TargetFrameworks>netstandard2.0;netcoreapp3.1;net5.0</TargetFrameworks>
 		<RootNamespace>Prometheus</RootNamespace>
 
 		<GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+		<GenerateDocumentationFile>true</GenerateDocumentationFile>
 
 		<SignAssembly>true</SignAssembly>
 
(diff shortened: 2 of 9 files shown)


Prometheus.AspNetCore.Grpc/Prometheus.AspNetCore.Grpc.csproj
```diff
@@ -1,10 +1,11 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
     <RootNamespace>Prometheus</RootNamespace>
 
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
baronfel, line 8: Adding this lets us remove the xml file customization below in favor of the more general, out of the box correct version.

Prometheus.AspNetCore.Grpc/Prometheus.AspNetCore.Grpc.csproj
```diff
@@ -16,22 +17,13 @@
...
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <DocumentationFile>bin\Release\netstandard2.0\Prometheus.Grpc.xml</DocumentationFile>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <DocumentationFile>bin\Debug\netstandard2.0\Prometheus.Grpc.xml</DocumentationFile>
+    <IsPackable>true</IsPackable>
+    <OutputType>Library</OutputType>
```
baronfel, line 21: Adding these two lines allows `dotnet pack` to work in conjunction with the web SDK.

Prometheus.AspNetCore.HealthChecks/Prometheus.AspNetCore.HealthChecks.csproj
```diff
@@ -14,33 +15,25 @@
 		<Nullable>enable</Nullable>
 		<WarningsAsErrors />
 		<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-	</PropertyGroup>
-
-	<PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-		<DocumentationFile>bin\Release\netstandard2.0\Prometheus.AspNetCore.HealthChecks.xml</DocumentationFile>
-		<CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
```
baronfel, line 21: these properties were the same across all configurations, so I moved them to the top level for clarity.

Prometheus.AspNetCore.HealthChecks/Prometheus.AspNetCore.HealthChecks.csproj
```diff
@@ -14,33 +15,25 @@
...
-		<FrameworkReference Include="Microsoft.AspNetCore.App" />
-	</ItemGroup>
-
 	<ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
 		<PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />
 	</ItemGroup>
 
-	<ItemGroup>
+	<ItemGroup Condition=" '$(TargetFramework)' == 'net5.0'">
+		<PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="5.0.5" />
```
baronfel, line 33: this dependency doesn't have .netcoreapp3.1 versions in the 5.x version line, so it needed to be conditionally-changed based on the TFM.

Prometheus.AspNetCore/HttpClientMetrics/HttpClientDelegatingHandlerBase.cs
```diff
@@ -68,7 +68,7 @@ protected internal TChild CreateChild(HttpRequestMessage request)
                         labelValues[i] = request.Method.Method;
                         break;
                     case HttpClientRequestLabelNames.Host:
-                        labelValues[i] = request.RequestUri.Host;
+                        labelValues[i] = request.RequestUri?.Host ?? string.Empty;
```
baronfel, line 71: is this a sensible default? opting in to nullability for net5.0 triggered a warning that `RequestUri` might be null.

Prometheus.AspNetCore/KestrelMetricServer.cs
```diff
@@ -56,7 +56,7 @@ protected override Task StartServer(CancellationToken cancel)
                     // If there is any URL prefix, we just redirect people going to root URL to our prefix.
                     if (!string.IsNullOrWhiteSpace(_url.Trim('/')))
                     {
-                        app.MapWhen(context => context.Request.Path.Value.Trim('/') == "",
+                        app.MapWhen(context => context.Request.Path.Value?.Trim('/') == "",
```
baronfel, line 59: similarly, in net5.0 `Value` may be null here.

Prometheus.AspNetCore/MetricServerMiddlewareExtensions.cs
```diff
@@ -8,7 +8,7 @@ namespace Prometheus
     public static class MetricServerMiddlewareExtensions
     {
 
-#if NETCOREAPP3_1
+#if NETCOREAPP3_1 || NET5_0
```
baronfel, line 11: this is needed in order to make this extension available on net5.0

Prometheus.NetStandard/Prometheus.NetStandard.csproj
```diff
@@ -32,6 +29,11 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+    <!-- Enables building, but not necessarily running on non-windows systems -->
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies.net461" Version="1.0.0">
```
baronfel, line 33: adding this metapackage allows us to build the project on mac/linux hosts, which helps contributors.

Prometheus.NetStandard/Prometheus.NetStandard.csproj
```diff
@@ -1,9 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup Condition=" '$(OS)' != 'Windows_NT' ">
```
baronfel, line 3: adding the reference assemblies package means we no longer have to condition the build based on OS.

CycoPH:
When is this being merged in?  I would love to use it in .net 5

sandersaares:
I believe this got sorted with a different fix described in #267. Closing this one.

Issue #284, opened 2021-01-08
Support OpenMetrics


joe-elliott:
The OpenMetrics standard has now been released.  Supporting this standard will bring prometheus-net up to current and allow users to make use of exemplars.

Please add support for this format to prometheus-net

https://github.com/OpenObservability/OpenMetrics/blob/master/specification/OpenMetrics.md

atheriel:
An list of changes that would need to be made for an MVP, from my recollection:

* Content-type support (including negotiation, which is what the Python and Go clients do at present).
* Enforce that counters end with `_total`. (The OpenMetrics name for a counter does not include the `_total` suffix, it has a `_total` child metric instead.)
* Add `_created` child metrics for counters (and histograms?).

Potential features of OpenMetrics once basic support is included:

* `UNIT` annotations.
* Additional metric families, including Info, StateSets, and GaugeHistograms.
* Exemplars (as mentioned), potentially the most exciting feature for users.

I am willing to take on some of the initial work if there's interest. I wrote another OpenMetrics client and have some experience.

chavalikesavanand:
Any update by when exemplar support be available?

PR #300, opened 2021-04-20
Add initial OpenMetrics support

atheriel:
As proposed in #284: the new [OpenMetrics](https://openmetrics.io/) standard is the next-gen format for consuming and emitting Prometheus metrics, and has been supported by Prometheus itself for a few years now.

This PR does the following to achieve basic compatibility but doesn't add any OpenMetrics-specific features:

* Make the serializer aware of the content type it is writing so it can handle different scenarios internally.
* Ensure the stream ends with `# EOF` when writing the OpenMetrics format (as required by the standard).
* Work around Counter incompatibilities in the OpenMetrics format.
* Serve the OpenMetrics format from the MetricServer and Asp.Net Core middleware via Content-Type negotiation.

There are tests for all new features included, and I have also tested that metrics emitted with these changes can be parsed by the [de-facto standard OpenMetrics parser](https://github.com/OpenObservability/OpenMetrics/blob/main/tests/implementations/prometheus_client_python_parser/parser.py) (part of the Prometheus Python client).

I'm happy to take criticism of the implementation, as well. As an aside: I have actually implemented most of the OpenMetrics features on top of these changes (including `_created`, units, and exemplars with automatic trace ID injection), so if there is interest I can attempt to upstream those as well.

cc @joe-elliott

### DIFF
--- a/Prometheus.AspNetCore/MetricServerMiddleware.cs
+++ b/Prometheus.AspNetCore/MetricServerMiddleware.cs
@@ -27,6 +27,8 @@ namespace Prometheus
         public async Task Invoke(HttpContext context)
         {
             var response = context.Response;
+            var contentType = context.Request.Headers["Accept"].ToString().Contains("application/openmetrics-text") ?
+                PrometheusConstants.ExporterContentTypeOpenMetrics : PrometheusConstants.ExporterContentType;
 
             try
             {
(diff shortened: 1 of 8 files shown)


jmajoor:
I would welcome full support for OpenMetrics. Prometheus also mentions to fully adapt the standard.
Definitely interested to see more OpenMetrics features too (including the protobuf format).

nicolastakashi:
Hi peeps, great work first of all.

Is there any idea to rollout it?

sandersaares:
We have recently merged an alternative implementation of OpenMetrics text format support by @hsyed-dojo. I believe this PR is now obsolete. Or is there any gap that might exist?

I would be interested to also hear more about future OpenMetrics integrations that may be relevant to your use cases (e.g. protobuf support).

atheriel:
I'm happy to hear that support has landed and I'll close this in response. I'm no longer using C# so I have little more to contribute at this point.

PR #302, opened 2021-04-28
Handle HttpClient timeout to avoid terminating push thread

### LINKED ISSUE #293: MetricsPusher can randomly terminate
We have seen this on some long-running applications. We start the metrics, but at some random point they die.

I looked at the code and found something I think may be a potential issue, that I've seen before in a different project:

In [MetricsPusher](https://github.com/prometheus-net/prometheus-net/blob/1cfbebb79719b9016284b7ded8dc5ef93e500fea/Prometheus.NetStandard/MetricPusher.cs) you use OperationCanceledException to check for cancelation due to cancelation token, however if the HTTPClient is configured with a timeout (it is in our case), then it can generate an OperationCanceledException without the CancelationToken being triggered, see [the official docs](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.postasync?view=net-5.0).

This would cause the loop to terminate silently, which would be a bug.

The solution is simply to check if the cancelation token is triggered, instead of relying on OperationCanceledException, which is a far safer and more correct way to do this to begin with. Even if my explanation is not correct, that should still be changed, since the loop should _never_ be allowed to terminate without an external cancelation.

runerys:
Fixes #293 

We've experienced production problems where the push thread terminated even though the host worker ran fine. 

The referenced issue #293 describes that the HttpClient, in case of a timeout will throw TaskCanceledException, a subtype of OperationCanceledException. We've recreated the error manually by setting HttpClient timeout to 1ms. This killed the push thread.

We can't see why exceptions inheriting OperationCanceledExceptions should be allowed to terminate the thread. The existing cancellation token handling seems sufficient.

This PR adds a fix in MetricPusher.cs + test scenarios that ensure that the exception types documented on PostAsync all results in a OnError callback. The timeout scenario yielded a breaking test before updating line 113 in MetricPusher.cs:
```c#
catch (Exception ex) // delete: when (!(ex is OperationCanceledException))
```

HttpClient PostAsync documentation: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.postasync?view=net-5.0

### DIFF
--- a/Prometheus.NetStandard/MetricPusher.cs
+++ b/Prometheus.NetStandard/MetricPusher.cs
@@ -110,7 +110,7 @@ namespace Prometheus
                         // We do not consider failed scrapes a reportable error since the user code that raises the failure should be the one logging it.
                         Trace.WriteLine($"Skipping metrics push due to failed scrape: {ex.Message}");
                     }
-                    catch (Exception ex) when (!(ex is OperationCanceledException))
+                    catch (Exception ex)
                     {
                         HandleFailedPush(ex);
                     }
--- a/Tests.NetFramework/MetricPusherTests.cs
+++ b/Tests.NetFramework/MetricPusherTests.cs
@@ -1,6 +1,8 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
+using System.Net.Http;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace Prometheus.Tests
 {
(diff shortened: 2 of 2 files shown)


yburke94:
We were also hit by this bug this week. @sandersaares can we get this change merged?

runerys:
@yburke94 This problem can be quick-fixed while waiting for a merge:
- Inherit the MetricHandler class and copy-paste the code with the fix into your own source. I called our "RobustMetricPusher"
`public class RobustMetricPusher : MetricHandler`
- You have to copy the following as well: DelegatingStreamInternal, PushStreamContentInternal and ValueStopWatch (they are already copied from .NET source)
- Use the new implementation instead of MetricPusher.

We have been running this fix in production for some time now, and I can confirm that it works.

mfkl:
How can we help move this forward?

sandersaares:
Your analysis makes sense. Thank you for the contribution!

PR #329, opened 2021-09-29
Reduce ASP.NET core metric cardinality when using `code`

djluck:
## Summary
Adding the option to reduce the cardinality of metrics tracked by prometheus-net by compressing the `code` label like so:
- `200` -> `2xx`
- `404` -> `4xx`

This can significantly reduce the cardinality of ASP.NET core count + latency metrics while still clearly communicating the difference between successfully processed requests, requests that failed due to client error and requests that failed due to a failure on the server.

This behavior defaults to disabled, allowing users to opt-in where they feel they can benefit.

## Testing
- Added a test to verify a small range of status codes can be mapped correctly.


sandersaares:
That's a great idea! Thank you for the contribution!

PR #331, opened 2021-10-22
Make the ASP.NET Exporter output the full value of the Content-Type header

Aegaina:


sandersaares:

PR #332, opened 2021-10-27
Allow custom routes via AspNetMetricServer

thestaphmonster:
This already looks to be possible with Prometheus.AspNetCore, but I couldn't see a simple way to do it with Prometheus.NetFramework.AspNet.


sandersaares:
Looks reasonable. Thank you for the contribution!

PR #353, opened 2022-04-01
Added MetricsUrl for HttpMetricsMiddleware

Isitar:
This extends the funcitonality for the CaptureMetricsUrl to specify another metrics url.


sandersaares:
I had some mixed feelings about the original implementation for exactly this reason - it only worked for the default URL - so it is good to see a PR on this topic. However, I also have mixed feelings about this approach to solve it - we put the burden on the user. Can we instead come up with a way to always do the right thing and just auto-detect the right URL or similar?

Isitar:
This implementation has a default to /metrics so everything works the same as before. More advanced use cases like mine can overwrite the url. 

What do you mean by auto detecting? In my use case i wanted it to be under /api/metrics, there is no real hint in the project that it should be hosted there.

PR #364, opened 2022-07-28
Use built in way to calculate unix time

JeremyMahieu:
Slightest imporvement possible but it saves a line and uses a standard way to convert it. Available in all the standards since Framework 4.6


sandersaares:
There is, in fact, a good reason `ToUnixTimeSeconds()` is not used - it loses the fractional part of the seconds! However, you are right that this can be simplified - we have `Gauge.SetToTimeUtc()` for exactly this purpose. I have manually merged the appropriate change - thank you for taking a poke at this!

JeremyMahieu:
well done, looks better than my suggestion

PR #365, opened 2022-07-28
Use IncTo and refactor foreach gen

JeremyMahieu:
Replaced weird usage of Inc() with IncTo() which seems to be made for this exact case.
And slight compaction of code with the foreach loop statement.

Sidenote, `GC.MaxGeneration` is designed to increase during runtime. If it does, it would throw each cycle. However in practice  `GC.MaxGeneration` is hardcoded to be 2. This would solve this (non-existing) problem however no new counters will be made for the newer generation that didn't exist at the start of the runtime.

### DIFF
--- a/Prometheus.NetStandard/DotNetStats.cs
+++ b/Prometheus.NetStandard/DotNetStats.cs
@@ -20,7 +20,7 @@ namespace Prometheus
         }
 
         private readonly Process _process;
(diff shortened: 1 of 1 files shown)


sandersaares:
Thank you for the contribution! I have manually merged the IncTo changes, though not the ForEach loop, as that felt a bit "different not better".

PR #369, opened 2022-08-19
Added benchmarks to prometheus-net README

kaffarell:
Made some simple benchmarks and thought they might fit in here...


kaffarell:
Going to add extensive benchmarks and graphs later today... This PR will stay the same though, so we can already merge this.

sandersaares:
This seems like material for an interesting blog post that I would love to read! However, it does not feel like suitably generic content for the readme, so I will close the PR without merging.

PR #385, opened 2022-11-10
Openmetrics refactor and tests

hsyed-dojo:
The core of this PR is about moving the business logic of the `CreateIdentity` encoding function to the `TextSerializer` 

The granularity of memoization has changed. The system labels `le `and `quantile` have corner cases when using the OpenMetrics encoding. For example le="1" in the Prom text format should be encoded as le="1.0" in OpenMetrics. 

I wrote some tests before undertaking the refactor, I will build on these when building out the OpenMetrics support.

Pre-refactor benchmark:

|              Method |     Mean |    Error |   StdDev |     Gen0 | Allocated |
|-------------------- |---------:|---------:|---------:|---------:|----------:|
| CollectAndSerialize | 40.07 ms | 0.520 ms | 0.461 ms | 692.3077 |   5.91 MB |

Refactored benchmark:

|              Method |     Mean |    Error |   StdDev |     Gen0 | Allocated |
|-------------------- |---------:|---------:|---------:|---------:|----------:|
| CollectAndSerialize | 40.53 ms | 0.375 ms | 0.333 ms | 692.3077 |   5.91 MB |

### DIFF
--- a/Prometheus/ChildBase.cs
+++ b/Prometheus/ChildBase.cs
@@ -7,7 +7,7 @@ namespace Prometheus
     {
         internal ChildBase(Collector parent, LabelSequence instanceLabels, LabelSequence flattenedLabels, bool publish)
         {
-            _parent = parent;
+            Parent = parent;
             InstanceLabels = instanceLabels;
             FlattenedLabels = flattenedLabels;
             FlattenedLabelsBytes = PrometheusConstants.ExportEncoding.GetBytes(flattenedLabels.Serialize());
@@ -44,7 +44,7 @@ namespace Prometheus
         /// </summary>
         public void Remove()
         {
-            _parent.RemoveLabelled(InstanceLabels);
+            Parent.RemoveLabelled(InstanceLabels);
         }
 
         public void Dispose() => Remove();
@@ -60,9 +60,10 @@ namespace Prometheus
         /// Internal for testing purposes only.
         /// </summary>
         internal LabelSequence FlattenedLabels { get; }
+
         internal byte[] FlattenedLabelsBytes { get; }
-        
-        internal readonly Collector _parent;         // TODO: rename to Parent (right?)
+
+        internal readonly Collector Parent;
 
         private bool _publish;
 
--- a/Prometheus/Summary.cs
+++ b/Prometheus/Summary.cs
@@ -92,20 +92,21 @@ namespace Prometheus
 
                 _headStream = _streams[0];
 
-                _quantileLabels = new Tuple<byte[], byte[]>[_objectives.Count];
+                _quantileLabels = new CanonicalLabel[_objectives.Count];
                 for (var i = 0; i < _objectives.Count; i++)
                 {
                     _sortedObjectives[i] = _objectives[i].Quantile;
-                    _quantileLabels[i] = TextSerializer.EncodeSystemLabelValue(_objectives[i].Quantile);
+                    _quantileLabels[i] = TextSerializer.EncodeValueAsCanonicalLabel(
+                        QuantileLabelName, _objectives[i].Quantile);
                 }
 
                 Array.Sort(_sortedObjectives);
             }
 
-            private readonly Tuple<byte[], byte[]>[] _quantileLabels;
-            private static readonly byte[] SumLabelBytes = PrometheusConstants.ExportEncoding.GetBytes("sum");
-            private static readonly byte[] CountLabelBytes = PrometheusConstants.ExportEncoding.GetBytes("count");
-            private static readonly byte[] QuantileLabelBytes = PrometheusConstants.ExportEncoding.GetBytes("quantile");
+            private readonly CanonicalLabel[] _quantileLabels;
+            private static readonly byte[] SumSuffix = PrometheusConstants.ExportEncoding.GetBytes("sum");
+            private static readonly byte[] CountSuffix = PrometheusConstants.ExportEncoding.GetBytes("count");
+            private static readonly byte[] QuantileLabelName = PrometheusConstants.ExportEncoding.GetBytes("quantile");
 
             private protected override async Task CollectAndSerializeImplAsync(IMetricsSerializer serializer,
                 CancellationToken cancel)
@@ -141,20 +142,29 @@ namespace Prometheus
                     }
                 }
 
-                await serializer.WriteIdentifierPartAsync(Parent.NameBytes, FlattenedLabelsBytes, cancel,
-                    postfix: SumLabelBytes);
-                await serializer.WriteValuePartAsync(sum, cancel);
-                await serializer.WriteIdentifierPartAsync(Parent.NameBytes, FlattenedLabelsBytes, cancel,
-                    postfix: CountLabelBytes);
-                await serializer.WriteValuePartAsync(count, cancel);
+                await serializer.WriteMetricPointAsync(
+                    Parent.NameBytes,
+                    FlattenedLabelsBytes,
+                    CanonicalLabel.Empty,
+                    cancel, 
+                    sum,
+                    suffix: SumSuffix);
+                await serializer.WriteMetricPointAsync(
+                    Parent.NameBytes,
+                    FlattenedLabelsBytes,
+                    CanonicalLabel.Empty,
+                    cancel, 
+                    count,
+                    suffix: CountSuffix);
 
                 for (var i = 0; i < values.Count; i++)
                 {
-                    await serializer.WriteIdentifierPartAsync(
-                        Parent.NameBytes, FlattenedLabelsBytes, cancel, postfix: null,
-                        extraLabelName: QuantileLabelBytes, extraLabelValue: _quantileLabels[i].Item1,
-                        extraLabelValueOpenMetrics: _quantileLabels[i].Item2);
-                    await serializer.WriteValuePartAsync(values[i].value, cancel);
+                    await serializer.WriteMetricPointAsync(
+                        Parent.NameBytes,
+                        FlattenedLabelsBytes,
+                        _quantileLabels[i],
+                        cancel, 
+                        values[i].value);
                 }
             }
 
--- a/Prometheus/Summary.cs
+++ b/Prometheus/Summary.cs
@@ -141,17 +141,17 @@ namespace Prometheus
                     }
                 }
 
-                await serializer.WriteIdentifierPartAsync(_parent.NameBytes, FlattenedLabelsBytes, cancel,
+                await serializer.WriteIdentifierPartAsync(Parent.NameBytes, FlattenedLabelsBytes, cancel,
                     postfix: SumLabelBytes);
                 await serializer.WriteValuePartAsync(sum, cancel);
-                await serializer.WriteIdentifierPartAsync(_parent.NameBytes, FlattenedLabelsBytes, cancel,
+                await serializer.WriteIdentifierPartAsync(Parent.NameBytes, FlattenedLabelsBytes, cancel,
                     postfix: CountLabelBytes);
                 await serializer.WriteValuePartAsync(count, cancel);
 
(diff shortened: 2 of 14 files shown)


Tests.NetCore/TextSerializerTests.cs
```diff
@@ -0,0 +1,197 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Prometheus.Tests;
+
+[TestClass]
+public class TextSerializerTests
```
sandersaares, line 10: The typical pattern we use is that every test class is also present in the Tests.NetFramework project via "include existing file", to ensure that the same logic also works under .NET Framework (if it is not .NET Core specific).

Tests.NetCore/TextSerializerTests.cs
```diff
@@ -0,0 +1,197 @@
...
+                LabelNames = new[] { "blah" },
+                Objectives = new[]
+                {
+                    new QuantileEpsilonPair(0.5, 0.05),
+                }
+            });
+
+            summary.WithLabels("foo").Observe(3);
+        });
+        // TODO help has a trailing whitespace before the newline
```
sandersaares, line 28: What are your thoughts on this? Planned for followup in next iterations, ignore for now? I would prefer to see some fix for it already. It is only help text but historically there have been apps taking a dependency on help text details and freaking out when it is unusual.

Prometheus/TextSerializer.cs
```diff
@@ -7,8 +7,15 @@ namespace Prometheus
...
-        private static readonly byte[] NewLine = new[] { (byte)'\n' };
-        private static readonly byte[] Space = new[] { (byte)' ' };
+        private static readonly byte[] NewLine = { (byte)'\n' };
+        private static readonly byte[] Quote = { (byte)'"' };
+        private static readonly byte[] Equal = { (byte)'=' };
+        private static readonly byte[] Comma = { (byte)',' };
+        private static readonly byte[] Underscore = { (byte)'_' };
+        private static readonly byte[] LBrace = { (byte)'{' };
+        private static readonly byte[] RBraceSp = { (byte)'}', (byte)' ' };
+        private static readonly byte[] Sp = { (byte)' ' };
```
sandersaares, line 17: I would prefer "Space" as everything else is nicely spelled out. [Just on a personal level, my brain expands Sp to Special]

Prometheus/TextSerializer.cs
```diff
@@ -7,8 +7,15 @@ namespace Prometheus
...
+        private static readonly byte[] PInf = PrometheusConstants.ExportEncoding.GetBytes("+Inf");
```
sandersaares, line 18: This might as well also be PositiveInfinity to be explicit, although it is more obvious than Sp

Prometheus/TextSerializer.cs
```diff
@@ -49,19 +56,77 @@ public async Task WriteFamilyDeclarationAsync(byte[][] headerLines, Cancellation
...
+            {
+                await _stream.Value.WriteAsync(Sp, 0, Sp.Length, cancel);
+            }
+        }
+
+        /// <summary>
+        /// Encode the system variable in regular Prometheus form and also return a OpenMetrics variant, these can be
+        /// the same.
+        /// </summary>
+        internal static Tuple<byte[], byte[]> EncodeSystemLabelValue(double value)
```
sandersaares, line 120: What does the word "system" here refer to? I was not able to find a definition of "system label" in OpenMetrics, for example. It is not clear what is the difference between "system label" and "<something else> label". Is there even a difference? Could use some clarification.

sandersaares, line 120: I would prefer some DTO instead of Tuple, so we clearly name every returned value. Or are there extra considerations that force Tuple to be preferred (e.g. perf)?

Prometheus/TextSerializer.cs
```diff
@@ -49,19 +56,77 @@ public async Task WriteFamilyDeclarationAsync(byte[][] headerLines, Cancellation
...
+        /// <summary>
+        /// Creates a metric identifier, with an optional name postfix and an optional extra label to append to the end.
+        /// familyname_postfix{labelkey1="labelvalue1",labelkey2="labelvalue2"}
+        /// Note: Terminates with a SPACE
+        /// </summary>
+        public async Task WriteIdentifierPartAsync(byte[] name, byte[] flatennedLabels, CancellationToken cancel,
+            byte[]? postfix = null, byte[]? extraLabelName = null, byte[]? extraLabelValue = null,
+            byte[]? extraLabelValueOpenMetrics = null)
+        {
+            await _stream.Value.WriteAsync(name, 0, name.Length, cancel);
```
sandersaares, line 80: There was an issue reported to prometheus-net repo a while ago that under ??? conditions the reporter saw that there was performance degradation from the many WriteAsync() calls. They never specified how to reproduce, though, so I do not know the exact situation.

There are now many more WriteAsync calls. I understand you ran the benchmarks but I wonder if there will be some problem that is not exposed by the benchmarks (e.g. because of MemoryStream versus network stream differences). Thoughts?

Prometheus/Summary.cs
```diff
@@ -96,30 +92,23 @@ internal Child(Summary parent, LabelSequence instanceLabels, LabelSequence flatt
...
-
-                    _quantileIdentifiers[i] = CreateIdentifier(null, "quantile", value);
-                }
             }
 
-            private readonly byte[] _sumIdentifier;
-            private readonly byte[] _countIdentifier;
-            private readonly byte[][] _quantileIdentifiers;
+            private readonly Tuple<byte[], byte[]>[] _quantileLabels;
+            private static readonly byte[] SumLabelBytes = PrometheusConstants.ExportEncoding.GetBytes("sum");
```
sandersaares, line 106: To me a "label" is the name-value pair, so I would expect to see both in a variable named `SumLabelBytes`. Perhaps `SumIdentifierBytes` or `SumNameBytes` is better if this is only the name of the label.

Prometheus/Summary.cs
```diff
@@ -152,11 +141,21 @@ private protected override async Task CollectAndSerializeImplAsync(IMetricsSeria
                     }
                 }
 
-                await serializer.WriteMetricAsync(_sumIdentifier, sum, cancel);
-                await serializer.WriteMetricAsync(_countIdentifier, count, cancel);
+                await serializer.WriteIdentifierPartAsync(Parent.NameBytes, FlattenedLabelsBytes, cancel,
```
sandersaares, line 144: Would it make sense to still keep it 1 call on this level and split into identifier/value parts inside the TextSerializer, just to keep the calling code simpler? I suppose this depends on what you have in mind for the next iteration. Ideally we can put all the complexity in the serializer but it could be I am thinking too simplistically here.

Prometheus/TextSerializer.cs
```diff
@@ -49,19 +56,77 @@ public async Task WriteFamilyDeclarationAsync(byte[][] headerLines, Cancellation
...
         }
+
+        /// <summary>
+        /// Creates a metric identifier, with an optional name postfix and an optional extra label to append to the end.
+        /// familyname_postfix{labelkey1="labelvalue1",labelkey2="labelvalue2"}
+        /// Note: Terminates with a SPACE
+        /// </summary>
+        public async Task WriteIdentifierPartAsync(byte[] name, byte[] flatennedLabels, CancellationToken cancel,
+            byte[]? postfix = null, byte[]? extraLabelName = null, byte[]? extraLabelValue = null,
+            byte[]? extraLabelValueOpenMetrics = null)
```
sandersaares, line 78: extraLabelValue for Prometheus and extraLabelValueOpenMetrics for OpenMetrics is the pattern, right? If so, I would prefer "extraLabelValuePrometheus" for the former, to treat the two equally.

Alternatively, can we provide some DTO here that contains both values in one argument? This special logic for the extra label here might be easier to understand that way.

Alternatively, can the caller already know which output format we use and only provide the right value?

Prometheus/ChildBase.cs
```diff
@@ -80,32 +83,5 @@ internal Task CollectAndSerializeAsync(IMetricsSerializer serializer, Cancellati
...
-            var labels = FlattenedLabels;
-
-            if (extraLabelName != null && extraLabelValue != null)
-            {
-                var extraLabelNames = StringSequence.From(extraLabelName);
-                var extraLabelValues = StringSequence.From(extraLabelValue);
-
-                var extraLabels = LabelSequence.From(extraLabelNames, extraLabelValues);
-
-                // Extra labels go to the end (i.e. they are deepest to inherit from).
```
sandersaares, line 101: This is likely a valuable comment to preserve in the logic that was moved to TextSerializer, to emphasize that the labels are in a specific order intentionally

sandersaares (changes requested):

Prometheus/TextSerializer.cs
hsyed-dojo: In the OpenMetrics spec [it is documented as "canonical numbers" ](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#considerations-canonical-numbers).
 
Renamed function accordingly and added link to openmetrics spec.

hsyed-dojo: I added a `CanonicalLabel` struct. It includes all three of the optional fields. I couldn't pass it by ref (async method) so had to add an `IsNonEmpty` method along with a sentinel.

Prometheus/TextSerializer.cs
hsyed-dojo: Added DTO 👍

Prometheus/Summary.cs
hsyed-dojo: I've canocalised the convention `*Suffix` and `*LabelName`.

Prometheus/Summary.cs
hsyed-dojo: Thank you for the push. I have added a `WriteMetricPointAsync` function and encapsulated the others away. OpenMetrics encoding concerns should fit that single function 🤞 .

Prometheus/TextSerializer.cs
hsyed-dojo: @sandersaares I too am concerned about this (ref: out of band discussion). 

I propose we tackle this towards the end of the feature. 

If the PR looks good perhaps we can have a staging branch ? I envisage two more PRs to feature completion.

1. OpenMetrics encoding support -- minimal, aiming for feature parity with golang -- with tests.
2. HTTP Protocol negotiation function and make OpenMetrics configurable as a protocol programatically.

Prometheus/TextSerializer.cs
```diff
@@ -43,25 +50,92 @@ public async Task WriteFamilyDeclarationAsync(byte[][] headerLines, Cancellation
...
             await _stream.Value.WriteAsync(NewLine, 0, NewLine.Length, cancel);
         }
+
+        /// <summary>
+        /// Creates a metric identifier, with an optional name postfix and an optional extra label to append to the end.
+        /// familyname_postfix{labelkey1="labelvalue1",labelkey2="labelvalue2"}
+        /// Note: Terminates with a SPACE
+        /// </summary>
+        private async Task WriteIdentifierPartAsync(byte[] name, byte[] flattenedLabels, CancellationToken cancel,
+            CanonicalLabel canonicalLabel, byte[]? suffix = null)
```
sandersaares, line 85: Suggested naming: canonicalLabel -> extraLabel

Prometheus/TextSerializer.cs
sandersaares: Looks good. Agree to worry about async at a later point. I will merge this PR and we can iterate, either directly into master branch or to a separate branch if the next iteration is not merge-ready yet.

PR #388, opened 2022-11-18
Openmetrics encoding support and Exemplars.

hsyed-dojo:
This PR:

* Introduces all of the machinery needed to attach an Exemplar to Counter and Histgram instruments.
* Openmetrics encoding - aim to have feature parity with the Golang implementation ([this PR](https://github.com/prometheus/common/commit/3888f9a53d4582c8b524af4acfdee6af64bedf07)).
* Protocol negotiation to select encoding format (manually tested with curl and verified with prometheus).
* Wired up end to end in ASP.net (please let me know if I have missed other ways of starting a HTTP runloop).
* Bash script to setup prom for scraping and ingesting exemplars for manual testing.

State: Feature complete.

### DIFF
--- a/Prometheus/Counter.cs
+++ b/Prometheus/Counter.cs
@@ -10,22 +10,30 @@ namespace Prometheus
             }
 
             private ThreadSafeDouble _value;
+            private ObservedExemplar _observedExemplar = new();
 
             private protected override async Task CollectAndSerializeImplAsync(IMetricsSerializer serializer, CancellationToken cancel)
             {
+                ObservedExemplar ex;
+                lock(this)
+                    ex = _observedExemplar;
+                
                 await serializer.WriteMetricPointAsync(
                     Parent.NameBytes,
                     FlattenedLabelsBytes,
                     CanonicalLabel.Empty,
                     cancel, 
-                    Value);
+                    Value,
+                    ex);
             }
 
-            public void Inc(double increment = 1.0)
+            public void Inc(double increment = 1.0, params Exemplar.LabelPair[] exemplar)
             {
                 if (increment < 0.0)
                     throw new ArgumentOutOfRangeException(nameof(increment), "Counter value cannot decrease.");
-
+                
+                if (exemplar.Length > 0)
+                    lock (this) _observedExemplar.Update(exemplar, increment);
                 _value.Add(increment);
                 Publish();
             }
@@ -38,6 +46,7 @@ namespace Prometheus
 
             public double Value => _value.Value;
         }
+        
 
         private protected override Child NewChild(LabelSequence instanceLabels, LabelSequence flattenedLabels, bool publish)
         {
@@ -56,6 +65,9 @@ namespace Prometheus
         public void Publish() => Unlabelled.Publish();
         public void Unpublish() => Unlabelled.Unpublish();
 
+        public void Inc(double increment = 1, params Exemplar.LabelPair[] exemplar) =>
+            Unlabelled.Inc(increment, exemplar);
+        
         internal override MetricType Type => MetricType.Counter;
 
         internal override int TimeseriesCount => ChildCount;
--- a/Prometheus/Counter.cs
+++ b/Prometheus/Counter.cs
@@ -27,6 +27,11 @@ namespace Prometheus
                     ex);
             }
 
+            public void Inc(params Exemplar.LabelPair[] exemplar)
+            {
+                Inc(increment: 1, exemplar: exemplar);
+            }
+
             public void Inc(double increment = 1.0, params Exemplar.LabelPair[] exemplar)
             {
                 if (increment < 0.0)
@@ -64,6 +69,11 @@ namespace Prometheus
 
         public void Publish() => Unlabelled.Publish();
         public void Unpublish() => Unlabelled.Unpublish();
+
+        public void Inc(params Exemplar.LabelPair[] exemplar)
+        {
+            Inc(increment: 1, exemplar: exemplar);
+        }
 
         public void Inc(double increment = 1, params Exemplar.LabelPair[] exemplar) =>
             Unlabelled.Inc(increment, exemplar);
--- a/Prometheus/Counter.cs
+++ b/Prometheus/Counter.cs
@@ -10,21 +10,30 @@ namespace Prometheus
             }
 
             private ThreadSafeDouble _value;
-            private ObservedExemplar _observedExemplar = new();
+            private ObservedExemplar _observedExemplar = ObservedExemplar.Empty;
 
-            private protected override async Task CollectAndSerializeImplAsync(IMetricsSerializer serializer, CancellationToken cancel)
+            private protected override async Task CollectAndSerializeImplAsync(IMetricsSerializer serializer,
+                CancellationToken cancel)
             {
-                ObservedExemplar ex;
-                lock(this)
-                    ex = _observedExemplar;
-                
+                // Borrow the current exemplar
+                ObservedExemplar cp =
+                    Interlocked.CompareExchange(ref _observedExemplar, ObservedExemplar.Empty, _observedExemplar);
+
                 await serializer.WriteMetricPointAsync(
                     Parent.NameBytes,
                     FlattenedLabelsBytes,
                     CanonicalLabel.Empty,
-                    cancel, 
+                    cancel,
                     Value,
-                    ex);
+                    cp);
+
+                if (cp != ObservedExemplar.Empty)
+                {
+                    // attempt to return the exemplar to the pool unless a new one has arrived.
+                    var prev = Interlocked.CompareExchange(ref _observedExemplar, cp, ObservedExemplar.Empty);
+                    if (prev != ObservedExemplar.Empty) // a new exemplar is present so we return ours back to the pool. 
+                        ObservedExemplar.ReturnPooled(cp);
+                }
             }
 
             public void Inc(params Exemplar.LabelPair[] exemplar)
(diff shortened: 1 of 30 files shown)


Prometheus/Exemplar.cs
```diff
@@ -0,0 +1,71 @@
...
+        internal byte[] KeyBytes { get; }
+        internal byte[] ValueBytes { get; }
+    }
+    
+    /// <summary>
+    /// Return an exemplar label key, this may be curried with a value to produce a LabelPair.
+    /// </summary>
+    public static LabelKey Key(string key)
+    {
+        var si = new StringInfo(key);
```
hsyed-dojo, line 59: RFC: Need to find a better way than this.

Prometheus/Histogram.cs
```diff
@@ -83,36 +89,47 @@ internal Child(Histogram parent, LabelSequence instanceLabels, LabelSequence fla
...
                     _bucketCounts.Sum(b => b.Value),
+                    ObservedExemplar.Empty,
                     suffix: CountSuffix);
 
                 var cumulativeCount = 0L;
 
                 for (var i = 0; i < _bucketCounts.Length; i++)
                 {
+                    ObservedExemplar cp;
+                    lock (_exemplars)
```
hsyed-dojo, line 108: RFC

Prometheus/TextSerializer.cs
```diff
@@ -15,16 +15,26 @@ internal sealed class TextSerializer : IMetricsSerializer
...
-
-        public TextSerializer(Stream stream)
+        private static readonly byte[] NegativeInfinity = PrometheusConstants.ExportEncoding.GetBytes("-Inf");
+        private static readonly byte[] NotANumber = PrometheusConstants.ExportEncoding.GetBytes("NaN");
+        private static readonly byte[] PositiveOne = PrometheusConstants.ExportEncoding.GetBytes("1.0");
+        private static readonly byte[] Zero = PrometheusConstants.ExportEncoding.GetBytes("0.0");
+        private static readonly byte[] NegativeOne = PrometheusConstants.ExportEncoding.GetBytes("-1.0");
+        private static readonly byte[] EofNewLine = PrometheusConstants.ExportEncoding.GetBytes("# EOF\n");
+
+        public TextSerializer(Stream stream, ExpositionFormat fmt = ExpositionFormat.Text)
```
hsyed-dojo, line 27: @sandersaares happy to refactor this now that this is fleshed out. I do need some guidance on the primitives to use.

Prometheus/Exemplar.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System.Globalization;
+
+namespace Prometheus;
+
+public static class Exemplar
```
hsyed-dojo, line 5: RFC: Helper types to ensure the Exemplar atoms are captured efficiently for encoding.

TODO:

* Add empty/whitespace check for label and value (verify with spec).

Prometheus/Counter.cs
```diff
@@ -10,22 +10,30 @@ internal Child(Collector parent, LabelSequence instanceLabels, LabelSequence fla
             }
```
hsyed-dojo, line 10: The following comment from the go implementation has not yet been addressed, looks like encoding of the metric family header lines will also need to  format aware in this PR: 

```
// - Counters are expected to have the `_total` suffix in their metric name. In
//   the output, the suffix will be truncated from the `# TYPE` and `# HELP`
//   line. A counter with a missing `_total` suffix is not an error. However,
//   its type will be set to `unknown` in that case to avoid invalid OpenMetrics
//   output.
```

Prometheus/Exemplar.cs
hsyed-dojo: This approach I took above is counting graphemes (see [this](https://www.meziantou.net/how-to-correctly-count-the-number-of-characters-of-a-string.htm)), so it is incorrect.

We need to count the "Runes", the runes are unicode scalars. Runes are only available in dotnet core 3 and above.

@sandersaares could we drop support for older frameworks ?

hsyed-dojo: Hmm, what if we only allowed only ASCII characters here until we drop support for older frameworks ? I can't see any reason for an exemplar to contain anything but alphanumeric characters. We would need to assert the encoded `byte[]` and source string have same length.

Prometheus/Histogram.cs
sandersaares: It looks correct, yeah - it performs a shallow copy and that's it.

However, I am not very keen on having locks in the hot path (`ObserveInternal()`) - the metric objects try to be very lock-free in general because things like HTTP servers can funnel a lot of requests (handled by many threads) into observations on the same metric instances. So far I have managed to keep all except for Summary lock-free and would ideally like to keep it so.

You can see the potential cost of multithreaded access in `MeasurementBenchmarks` - already without locks it is rather slow (though artificially slow as there is no "other work" done in the benchmark). I am afraid that adding locks might create even more potential slowdown.

Example of differences on my PC (True/False column indicating use of exemplars; I will commit my exemplar-enhanced benchmark code shortly):

```
| MeasurementPerformance |           100000 |          16 |          Counter |         False | 124,351.9 us | 2,163.73 us | 2,023.96 us | 123,847.7 us |                    - |          13.0000 |     384 B |
| MeasurementPerformance |           100000 |          16 |          Counter |          True | 230,592.2 us | 3,085.85 us | 2,886.51 us | 229,846.1 us |                    - |        6124.0000 |     384 B |
|
| MeasurementPerformance |           100000 |          16 |        Histogram |         False | 212,602.3 us | 1,505.49 us | 1,408.24 us | 212,393.8 us |                    - |          14.0000 |     384 B |
| MeasurementPerformance |           100000 |          16 |        Histogram |          True | 242,188.5 us | 2,762.03 us | 2,583.61 us | 241,092.3 us |                    - |        1445.0000 |     384 B |
|
```

The total duration here, of course, also includes `ObservedExemplar.Update()` and is not only the lock contention overhead. The 2nd column from the right is the lock contention count.

A lock-free approach would be preferable here to try ensure that capturing data is cheap. How to do that, though? What if we turn it into a reference type and just pool them to avoid the constant allocation overhead?

* [ObjectPool](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.objectpool.objectpool.create?view=dotnet-plat-ext-7.0#microsoft-extensions-objectpool-objectpool-create-1(microsoft-extensions-objectpool-ipooledobjectpolicy((-0)))) can manage a set of `ObservedExemplar` instances.
* In `ObserveInternal()` we get a new instance from the pool and use a loop over [Interlocked.CompareExchange](https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked.compareexchange?view=net-7.0#system-threading-interlocked-compareexchange-1(-0@-0-0)) to replace it in `_exemplars`, returning the old one to the pool.
* In `CollectAndSerializeImplAsync()` we have a bit of a roadblock that we need to prevent our instance from being returned to the pool while we are still using it. However, we can perhaps do a mirrored operation, take an instance from the pool, zero it, and swap it out with the one in `_exemplars`. After serializing the exemplar, we can swap it back (if the one in `_exemplars` is still our empty instance) and return the empty instance to the pool.
* This leaves a small window where the exemplar is left empty, so in a theoretical case two concurrent metrics exports could end up with one not seeing the exemplar. But this seems like a totally OK price to pay for lock-free code - very theoretical situation, and harmless even if it happens.

What do you think? Just off the top of my head, it could be there is a big plot hole here somewhere.

Prometheus/ObservedExemplar.cs
```diff
@@ -0,0 +1,67 @@
...
+            for (var j = 0; j < labels.Length; j++)
+            {
+                if (i == j) continue;
+                if (Equal(labels[i].KeyBytes, labels[j].KeyBytes))
+                    throw new ArgumentException("exemplar contains duplicate keys");
+            }
+        }
+
+        if (tally > MaxRunes)
+            throw new ArgumentException($"exemplar labels has {tally} runes, exceeding the limit of {MaxRunes}.");
```
sandersaares, line 54: I am not very comfortable throwing "bad data" type of exception from this code, largely because it is often unexpected to the users of the library - they expect that telemetry APIs are not going to complain about data, it is seen as "harmless" code.

What is the impact if we skip the length check and emit over-length data?

Prometheus/Histogram.cs
```diff
@@ -124,6 +141,9 @@ public void Observe(double val, long count)
                     if (val <= _upperBounds[i])
                     {
                         _bucketCounts[i].Add(count);
+                        if (exemplar.Length > 0)
```
sandersaares, line 144: I suggest null-checking exemplar, to avoid surprising callers. Here and elsewhere, so we treat an exemplar of null as equivalent to empty array, to be more tolerant of the form of input that is provided.

Perhaps just `exemplar ??= Array.Empty<...>()` at the start of the method or such?

sandersaares (changes requested):

Prometheus/ObservedExemplar.cs
hsyed-dojo: [The go implementation has a panic in the instrumentation codepath.](https://github.com/prometheus/client_golang/blob/07b1397deda7a534e8fb59598c6cd6226038ff09/prometheus/histogram.go#L1008-L1019) I feel we should treat it similarly here.

> What is the impact if we skip the length check and emit over-length data?

I Suspect warnings from the storage backends in the logs and the Exemplars just being dropped.

Prometheus/ObservedExemplar.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        Labels = labels;
+        Val = val;
+        Timestamp = NowProvider.Now();
+    }
+    
+    private static bool Equal(byte[] a, byte[] b)
```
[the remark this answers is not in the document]
hsyed-dojo: I borrowed it from [here](https://www.syncfusion.com/succinctly-free-ebooks/application-security-in-net-succinctly/comparing-byte-arrays). Will add a comment.

Prometheus/Exemplar.cs
```diff
@@ -0,0 +1,74 @@
...
+        internal byte[] KeyBytes { get; }
+        internal byte[] ValueBytes { get; }
+    }
+    
+    /// <summary>
+    /// Return an exemplar label key, this may be curried with a value to produce a LabelPair.
+    /// </summary>
+    public static LabelKey Key(string key)
+    {
+        if (string.IsNullOrEmpty(key))
```
hsyed-dojo, line 59: `IsNullOrWhitespace`

Prometheus/Histogram.cs
hsyed-dojo: Rider suggested `if (exemplar is { Length: > 0 })` for a null + length check. That works (I think) ?

Prometheus/Exemplar.cs
hsyed-dojo: @sandersaares went with your proposed approach.

PR #235, opened 2020-08-18
Add gRPC request duration histogram

Sverdel:
I think, it would be nice to have duration histogram for grpc request just like for http requests. Unfortunately, `HttpRequestDurationMiddleware` can't be used, because it know nothing about gRPC method name. So I add `GrpcRequestDurationMiddleware` and all infrastructure for it

PR #432, opened 2023-08-09
Add histogram metrics for gRPC requests

EraYaN:
This is a continuation of #235 but this one is up to date.


EraYaN:
This has been rebased to latest master.

benjaminstammen:
Is there a timeline to getting this / something like this merged into master and new builds? Looking for this behavior in version 8.x so that I can avoid writing my own latency metrics.

PR #443, opened 2023-10-20
Perf meteradapter

exyi:


exyi:
The main remaining bottleneck now is the RWLock wrapping the LifetimeManager, in a profile 88% time is spent entering and exiting the lock. It's not that trivial to remove it, but I have an idea how to make this also lockfree (and also do away with the race which now may occur when deleting and incrementing metric at the same time). Would you be interested in that?

exyi:
Great! Moving the the lifetime timer higher up is definitely good news. At least for me it doesn't really matter if it's global or per-template metric. I don't have thousands of meters, just many different label values.

I rebased it on the current master branch, there was only a minor conflict. Do you have these changes in a private branch?

sandersaares:
Yeah, I am currently working in the "optimizing" branch

sandersaares:
```
| Method         | MeasurementCount | Mean     | Error    | StdDev   | Allocated |
|--------------- |----------------- |---------:|---------:|---------:|----------:|
| CounterInt     | 100000           | 31.15 ms | 0.336 ms | 0.314 ms |     106 B |
| CounterFloat   | 100000           | 31.03 ms | 0.283 ms | 0.265 ms |     106 B |
| HistogramInt   | 100000           | 32.06 ms | 0.169 ms | 0.158 ms |     106 B |
| HistogramFloat | 100000           | 31.52 ms | 0.296 ms | 0.277 ms |      53 B |
```

Okay that is starting to look a bit better. Thanks for peeking under the covers here, you motivated some good improvement both with your PR and with other ideas that came from this!

PR #447, opened 2023-11-11
Update Sdk comparison benchmarks

utpilla:
## Changes
- Use targeted `GlobalSetup` and `GlobalCleanup` instead of setting up and the disposing the SDK on every iteration of each benchmark method. This was unnecessary.
- Remove code to create duplicate Counters and Histogram This was just against the expected usage of these instruments.
- Removed benchmark to measure the setup cost. For OpenTelemetry, you cannot repeatedly create the same instrument. That's against the expected usage. You would need a more targeted benchmark for this. It's probably best to do that in a separate class.
- Add a new Benchmark method to measure how the different SDKs perform when the user gives a new unique dimension for every observation. This is added to mimic the scenario of tracking SLIs where you have to report a high number of unique dimension values such as `CustomerId`.
- Create and start a server in the PromNet setup as the OpenTelemetry setup also starts an HttpListener. This doesn't affect the metrics record benchmark numbers but it mimics a real-world use case better and offers a fair comparison for allocation during setup etc.
- Update OpenTelemetry Histogram to not record Min and Max as PromNet doesn't track that
- Update package versions of `BenchmarkDotNet` and `OpenTelemetry.Exporter.Prometheus.HttpListener`

## Benchmark Results

1. The first four benchmark methods record 1000 observations for each of the 100 time-series. In total each of these methods measure the performance of both the SDKs for reporting 100*1000 = 100,000 measurements. Both PromNet and OpenTelemetry SDKs are comparable here in terms of memory allocated. PromNet performs way better in terms of time taken.
2.  The last two benchmark methods for mimicking the SLI scenario is where OpenTelemetry outshines PromNet in both time taken, and memory allocated.

BenchmarkDotNet v0.13.10, Windows 11 (10.0.23424.1000)
Intel Core i7-9700 CPU 3.00GHz, 1 CPU, 8 logical and 8 physical cores
.NET SDK 8.0.100-rc.2.23502.2
  [Host]     : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2
  DefaultJob : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2
  Job-AGCLMW : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2

| Method                        | Job        | MaxIterationCount | Mean        | Error       | StdDev      | Gen0     | Gen1     | Allocated |
|------------------------------ |----------- |------------------ |------------:|------------:|------------:|---------:|---------:|----------:|
| PromNetCounter                | DefaultJob | Default           |    771.5 us |     5.54 us |     4.91 us |        - |        - |       1 B |
| PromNetHistogram              | DefaultJob | Default           |  2,747.5 us |    27.86 us |    26.06 us |        - |        - |       3 B |
| OTelCounter                   | DefaultJob | Default           | 14,470.8 us |    54.28 us |    48.12 us |        - |        - |      12 B |
| OTelHistogram                 | DefaultJob | Default           | 15,856.9 us |   193.51 us |   181.01 us |        - |        - |      25 B |
| PromNetHistogramForAdHocLabel | Job-AGCLMW | 16                |  8,804.0 us | 1,083.49 us | 1,013.49 us | 500.0000 | 234.3750 | 3184062 B |
| OTelHistogramForAdHocLabel    | Job-AGCLMW | 16                |    580.8 us |     6.08 us |     5.69 us |  14.6484 |        - |   96001 B |

### DIFF
--- a/Benchmark.NetCore/Benchmark.NetCore.csproj
+++ b/Benchmark.NetCore/Benchmark.NetCore.csproj
@@ -24,9 +24,9 @@
 	</ItemGroup>
 
 	<ItemGroup>
-		<PackageReference Include="BenchmarkDotNet" Version="0.13.2" />
+		<PackageReference Include="BenchmarkDotNet" Version="0.13.10" />
 		<PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="6.0.10" />
-		<PackageReference Include="OpenTelemetry.Exporter.Prometheus.AspNetCore" Version="1.4.0-rc.2" />
+		<PackageReference Include="OpenTelemetry.Exporter.Prometheus.HttpListener" Version="1.6.0-rc.1" />
 	</ItemGroup>
 
 	<ItemGroup>
--- a/Benchmark.NetCore/SdkComparisonBenchmarks.cs
+++ b/Benchmark.NetCore/SdkComparisonBenchmarks.cs
@@ -1,10 +1,28 @@
-﻿using System.Diagnostics;
-using System.Diagnostics.Metrics;
+﻿using System.Diagnostics.Metrics;
 using BenchmarkDotNet.Attributes;
 using OpenTelemetry.Metrics;
 using Prometheus;
 
 namespace Benchmark.NetCore;
+
+/*
+BenchmarkDotNet v0.13.10, Windows 11 (10.0.23424.1000)
+Intel Core i7-9700 CPU 3.00GHz, 1 CPU, 8 logical and 8 physical cores
+.NET SDK 8.0.100-rc.2.23502.2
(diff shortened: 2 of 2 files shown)


sandersaares:
> Remove code to create duplicate Counters and Histogram This was just against the expected usage of these instruments.

The purpose of this code was to aim for some realistic counts of metrics. In other words, it is not typical for apps to only have 1 counter and 1 histogram, so in terms of trying to achieve some semi-realistic workload, the benchmark used 100 of each.

That being said, it mostly just seems to apply a x100 multiplier to the values without showing any interesting differences, so I suppose it is ultimately irrelevant and fine to remove.

PR #448, opened 2023-11-18
feat: add path url http client metric

EngRajabi:
This feature https://github.com/prometheus-net/prometheus-net#ihttpclientfactory-metrics is a very good feature.
But it is very general, we may have 50 different api types from that client, and definitely not all of them are important! It may be necessary to set an alert service on a specific metric.
I added the path of the address so that we have statistics for each address


EngRajabi:
@sandersaares please check

EngRajabi:
@sandersaares please check

EngRajabi (commented):
please check

EngRajabi:
please check

PR #457, opened 2023-12-03
Cache strings in TextSerializer

dv00d00:
Not sure if intended but noticed this pattern:

`internal static ReadOnlySpan<byte> NewlineHashTypeSpace => [(byte)'\n', (byte)'#', (byte)' ', (byte)'T', (byte)'Y', (byte)'P', (byte)'E', (byte)' '];`

which is compiled to 

```
internal static ReadOnlySpan<byte> NewlineHashTypeSpace 
{
    get
    {
        return [(byte)'\n', (byte)'#', (byte)' ', (byte)'T', (byte)'Y', (byte)'P', (byte)'E', (byte)' '];
    }
}
```

Not sure if intended but better to be safe to ask are we really need new array every time, or maybe it is allocated on stack this way?

### DIFF
--- a/Benchmark.NetCore/AspNetCoreExporterBenchmarks.cs
+++ b/Benchmark.NetCore/AspNetCoreExporterBenchmarks.cs
@@ -44,6 +44,7 @@ public class AspNetCoreExporterBenchmarks
 
     private sealed class EntryPoint
     {
+#pragma warning disable CA1822 // Mark members as static
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddRouting();
@@ -66,6 +67,7 @@ public class AspNetCoreExporterBenchmarks
                 });
             });
         }
+#pragma warning restore CA1822 // Mark members as static
     }
 
     [Benchmark]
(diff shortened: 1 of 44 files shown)


sandersaares:
This pattern compiles down to a constant value, there is no allocation happening.

dv00d00:
| Method   | Runtime  | Mean      | Error     | StdDev    | Ratio | RatioSD | Allocated | Alloc Ratio |
|--------- |--------- |----------:|----------:|----------:|------:|--------:|----------:|------------:|
| Field    | .NET 6.0 | 3.1095 ns | 0.1513 ns | 0.0083 ns |  1.11 |    0.00 |         - |          NA |
| Property | .NET 6.0 | 2.8114 ns | 0.2281 ns | 0.0125 ns |  1.00 |    0.00 |         - |          NA |
|          |          |           |           |           |       |         |           |             |
| Field    | .NET 7.0 | 2.5297 ns | 0.1636 ns | 0.0090 ns |  0.92 |    0.00 |         - |          NA |
| Property | .NET 7.0 | 2.7620 ns | 0.2232 ns | 0.0122 ns |  1.00 |    0.00 |         - |          NA |
|          |          |           |           |           |       |         |           |             |
| Field    | .NET 8.0 | 2.3478 ns | 0.8396 ns | 0.0460 ns |  4.84 |    0.08 |         - |          NA |
| Property | .NET 8.0 | 0.4852 ns | 0.0881 ns | 0.0048 ns |  1.00 |    0.00 |         - |          NA |

PR #489, opened 2024-11-22
feat: Use github actions to run tests

jetelain:
Github actions:
- Linux for NetCore (prometheus-net-linux.slnf)
- Windows for NetFramework

Opted for PackageReference instead of packages.config to fix build with msbuild. Binding redirects will have to be updated manually as Visual Studio tooling is not able to update them with PackageReference.

Some dependencies have been updated to fix critical vulnerabilities.


jetelain:
The AzureDevOps pipeline is not compatible with PackageReference in net framework csproj (non sdk style).
An option would to merge Tests.NetCore and Tests.NetFramework in a single multi-target sdk style project.

